Issue #5, opened 2020-09-09
Terminology Changes 


laurie-tyz:
"Work Item":
 - Review the various places the phrase is used
 - implement consistently
 - a work item may be something besides a 'patch'.  It could be a mitigation, accepting risk, or verifying product is at end of life.
 - [x] Update usage of "Work Item" to align with the above. 
 

For both of "Patch Applier" and "Patch Developer," the following considerations apply: 
 - actor description implies only binary actions. Implementing a mitigation, accepting risk or acknowledging product is EOL.
 - Suggest and discuss new ROLE identifiers.  Eg., system owner, consumer, defender, deployer, supplier, provider, or distributor.
 - Substitute the new ROLE identifier throughout the document.
 - Review the changes for awkward, or unclear language.
 - [x] change "Patch Applier" to "Supplier"
 - [x] change "Patch Developer" to "Deployer"

"Worst credible" 
- [x] substitute for "worst plausible"

"Out-of-Cycle"
- [x] substitute for "out-of-band" in document
- [x] substitute for "out-of-band" in trees and code that produces trees (@j--- )

ahouseholder:
There's room for discussion, but I've been using "system owner" as the stand-in term for the individual or organization that controls the plugged-in boxes and has an interest/skin-in-the-game for the security of the deployed system. System owners might delegate some of those responsibilities to others (patch appliers, sysadmins, security guards, service providers, etc.) but the buck stops at the system owner I think.

ahouseholder:
@laurie-tyz has the lead on this issue.

laurie-tyz:
The term 'work item' is gone.  Replacement with the use of Work or Work unit. 

I will close out this item and begin working on Issue #46

PR #37, opened 2020-09-28
Issue #5

laurie-tyz:
changes complete:
s/worst **plausible**/worst **credible**/g **
s/out-of-band/out-of-cycle/g

** "worst" not located in the document. 

please void the merge if my work is in error.


doc/version_1/040_treesForVulMgmt.md
```diff
@@ -192,7 +191,7 @@ Table 7: Utility to the Adversary, as a Combination of Virulence and Value Densi
...
+Alternative heuristics for proxying adversary utility are credible. One such example is the value the vulnerability would have were it sold on the open market. Some firms, such as [Zerodium](https://zerodium.com/program.html), make such pricing structures public. The valuable exploits track the virulence and value density heuristics for the most part. Within a single system—whether it is Apache, Windows, iOS or WhatsApp—more automated kill chain steps successfully leads to higher exploit value. Remote code execution with sandbox escape and without user interaction are the most valuable exploits, and those features describe automation of the relevant kill chain steps. How equivalently virulent exploits for different systems are priced relative to each other is more idiosyncratic. Price does not only track value density of the system, but presumably also the existing supply of exploits and the installation distribution among the targets of Zerodium’s customers. Currently, we simplify the analysis and ignore these factors. However, future work should look for and prevent large mismatches between the outputs of the *utility* decision point and the exploit markets.
```
j---, line 194: My memory is that this line is not using "plausible" in the sense of reasoning about a worst-plausible result of an exploit. "Plausible" here is just introducing that the authors might make some other choices in the future about how the decision point is evaluated. So I think keeping it as "plausible" here would make this difference clearer, there was no way to tell in the previous text that this use of "plausible" was different than the earlier ones.

j--- (commented):
One place where the original text was unclear and I'm not sure a substitution is what we want.

doc/version_1/040_treesForVulMgmt.md
laurie-tyz: I agree.  That was one substitution I suspected was unintended.  
Thank you

Issue #9, opened 2020-09-09
"Virulence" is not an obviously-meaningful term 


j---:
Consider renaming *Virulence* to something more evocative / descriptive, like "wormable." If *Virulence* remains the title, consider listing synonyms in the heading or foregrounding the common-sense intuition of what it means.

ahouseholder:
Since we're borrowing terms from epidemiology, there are a few related terms that might help us figure this out:

- _virulence_ ["a pathogen's or microbe's ability to infect or damage a host"](https://en.wikipedia.org/wiki/Virulence)
- _infectivity_ ["the ability of a pathogen to establish an infection"](https://en.wikipedia.org/wiki/Infectivity) 
- _transmissibility_  is more about the spread of a pathogen between individuals https://en.wikipedia.org/wiki/Transmission_(medicine)

I think our intended meaning was less about virulence and more about infectivity or maybe transmissibility.

ahouseholder:
_automatability_ might be close.

The closely related (to each other) concepts of 
- _combustibility_ propensity to burst into flame if exposed to the right conditions, i.e., heat
- _flammability_ propensity to ignite easily at ambient temperatures
are analogs to the thing I think we're trying to represent here too.

Also relevant: from [CVSS v3.1 ](https://www.first.org/cvss/specification-document)
- _Attack Vector (AV)_ (Network/Adjacent/Local/Physical)
- _Attack Complexity (AC)_ (Low/High)
- _Privileges Required (PR)_ (None / Low / High)
- _User Interaction (UI)_ (None / Required)

Seems like in support of #33 and #30 we ought to map the CVSS stuff into _slow/rapid_ or whatever the end result is.

PR #40, opened 2020-10-07
s/virulence/automatability/g

ahouseholder:
Proposed fix for #9 

We had already used the word "automatable" in the definitions of what rapid and slow virulence meant. So I chose _automatability_ as the replacement term.

_transmissibility_ is the closest concept from epidemiology, I think. But it presumes a more natural context where transmission just stochastically happens or not, whereas we're trying to convey something about the likelihood of achieving an engineering goal for fully automated rapid exploitation.

_wormable_ is (a) ugly and (b) misses a bit where it may be possible to have a high degree of automation yet not be fully self-propagating code. Examples of "rapid automation yet not really a worm" would include attacks that propagate rapidly through social media once a user interacts with an object (e.g., click on an image, and suddenly you just posted it to all your connections).


j--- (approved):
Agree, I think this makes sense. 
Hopefully this is clear to analysts as well, we'll want to do a bit of testing there.

Issue #16, opened 2020-09-17
Clarify kill chain description in virulence


j---:
Make clear what is being automated by whom in the definition of virulence (or whatever it is called in v2, #9 )

PR #41, opened 2020-10-07
rephrase kill chain automation text to active voice

ahouseholder:
Proposed fix for #16, just converting to active voice to clarify slow vs rapid virulence/automatability


j--- (commented):
Looks good to me. If someone dislikes "attacker" as the term, they can open a new issue.

PR #42, opened 2020-10-08
Decision trees

laurie-tyz:
I've updated the following

s/patch applier/remediator/g
s/patch developer/supplier/g

removed "patch" from the entire section.

Please review and approve or reject.  We may need an additional meeting with Art to finalize his suggestion.


zmanion:
Closing, likely choosing "deployer" vs. "remediator"

Issue #35, opened 2020-09-22
Cost of mitigation


j---:
Section 4 of v1.1 says:
 > There are at least three aspects of asset management that may be important but are out of scope for SSVC. First and most obvious is the transaction cost of conducting the mitigation or fix. System administrators are paid to develop or apply any fixes or mitigations, and there may be other transactional costs such as downtime for updates. Second is the risk of the fix or mitigation introducing a new error or vulnerability. Regression testing is part of managing this type of risk. Finally, there may be an operational cost of applying a fix or mitigation, representing an ongoing change of functionality or increased overhead.

These three costs (transaction, error introduction, and operational change) are listed as out of scope in v1.1 under the assumption that these costs should not change the priority with which a vul ought to be patched. This assumption likely over-simplifies the vulnerability management decision. Version 2 should enumerate the situations in which these costs may legitimately delay the priority with which a vulnerability should be handled. If these situations are both realistic and justified (which seems likely), then how and why these considerations influence the vul handling decision should be integrated into SSVC. 

This is important for #14, for example.

j---:
One way we might handle this is to convert "mission impact" to "net mission impact" where "net" signals that we balance cost of acting with the cost of not acting. But unclear if that will be a usable solution.

zmanion:
See 3.6 in published SSVC and 4.5(?) in v2 draft.

j---:
See also https://www.scadafence.com/public-preview-a-comprehensive-guide-to-industrial-device-patching/

zmanion:
After some consideration/discussion:

SSVC recommends a priority with which to act. This priority approximates the cost/risk of not acting, i.e., higher priority ~= higher cost/risk to not acting, i.e., you are more likely to be attacked and incur/cause losses. @j--- to integrate this framing appropriately throughout v2 paper (including near the front).

SSVC does not address your cost/risk of acting, that is out of (the current) scope and is a separate/next decision to be made. Armed with the SSVC priority (== cost/risk of not acting), compare that to your cost/risk of acting (possibly for different actions), and make this next decision (which action to take).

SSVC "defer" is equivalent to the cost/risk of not acting being so low that it is not necessary to consider taking any other action, not necessary to consider the (low) cost/risk of not acting to the cost/risk of acting, not acting is always lower. Other SSVC priorities imply high enough cost/risk of not acting that you need to advance to the second decision and decide an action to take. It's still possible to choose to not act at this point, but we're beyond SSVC now.

[...]

PR #43, opened 2020-10-08
add paragraph about risk tolerance

### LINKED ISSUE #20: Discuss how risk posture or risk appetite influences labels on trees
For example, *immediate* maybe should be reachable with *Exploit* = none? If not the default recommendation, explain why and under what conditions it might make sense for someone to adopt this risk posture.

ahouseholder:
fix #20


j--- (changes requested):
This is good.
Is there any citation or other document that discusses the different categories of risks that need to be balanced in infosec? 
OCTAVE does not have the level of categorization I'm hoping for. 

Do we have a definition of "risk" that we agree with? 
I'd like to avoid the simple (impact x likelihood) formulation of risk, not that it's wrong per se, but because it leaves out distribution of harms, correlated risks, and the uncertainty distribution on the likelihood. 

Should both of these be new issues, or are they necessary for closing this one?

ahouseholder:
I think the risk definition question is a separate issue. See #49 

Categories of risk: I'm not familiar with any canonical lists, but can take a look.

ahouseholder:
This might be relevant:
Cebula, James., & Young, Lisa. 2010. A Taxonomy of Operational Cyber Security Risks (Technical Report CMU/SEI-2010-TN-028). Pittsburgh: Software Engineering Institute, Carnegie Mellon University. http://resources.sei.cmu.edu/library/asset-view.cfm?AssetID=9395

It seems to me that "Change risk" is a combination of subclass 2.2 ("systems and technology failures:software") and class 3 ("Failed internal processes") -- there is risk arising from making a change to an operational context that might not be able to absorb it well.

Whereas "Vulnerability risk" is a type of 1.2 ("actions of people:deliberate") -- the risk arises from an attacker intentionally causing harm.

If that makes sense I can modify the added text to refer to those.

j---:
I think the paper by Jim and Lisa is good. I would suggest that "change risk" includes all of 2 and 3, not just 2.2 and 3. The change might overload some hardware (2.1) or cause an integration failure (2.3.3), for example. But overall, I like it, and the document links to other standards as well, so people can map to their standard if there is one they must comply with. So I'm in favor of adding that in. 

We should make sure that we don't lose sight of cost of doing the work, but that's #35 and can be handled there. We might want to leave some breadcrumbs to make sure we cross reference these two discussions with each other in the document, is all.

ahouseholder:
Ok, added para on line 73 of diff to refer to the Risk Taxonomy paper. Citation added to .bib file too.

j--- (approved):
Thanks! Looks good to me.

PR #54, opened 2020-10-27
Address EOL products in scope section

### LINKED ISSUE #13: Status of EOL software
Clarify that EOL software is out of scope, because those vendors won't be using the tree or any vul mgmt. Appliers who own EOL'd software should still do vul mgmt, their "work item" just can no longer be "apply a patch."

ahouseholder:
Fix #13 

Adds a paragraph to scope giving an example of how suppliers and deployers might have different positions on whether EOL software is in or out of scope.


j--- (approved):
I like the way this handles the issue.

PR #63, opened 2020-11-09
Fix22

### LINKED ISSUE #22: Tree size and CSV documentation mode
How many rows in the CSV file that exhaustively defines a tree are too many? When does the human decision-maker (who has to approve the risk posture of the tree #20 ) lose track of the context? Discuss aspects of how many distinct combinations of decision values are too many.

j---:
The file for section 3 is a bit messy (as far as diffs) because I changed the line breaking to be one sentence per line rather than one paragraph per line. 

But the only substantive changes start at line 125 of that file. 

The goal is for this to fix #22


ahouseholder (commented):
This looks good for now. I don't have any reason to say that either 10 or 20 situations are better than 15 as a cutoff point, so I'm not strongly committed to the 60 unique combinations being the threshhold for "too big", but I think it's better to have a number as an anchor point that we might revise later than it is to have no anchor at all.

Issue #65, opened 2020-11-10
Full format communication


j---:
In subsection Guidance on Communicating Results (Section 4) determine what we should do about full format result formatting. It may be too early to do this well, so it may not be appropriate for v2. 

This is closely related to #24 and more specifically #25 #26 and #29.

sei-vsarvepalli:
Jono,

I think this is not difficult to do (unless I understood the request wrong). I developed one such representation for the SSVC calculator tool I wrote up, which basically can represent a full decision tree 

Schema - https://github.com/CERTCC/SSVC/blob/main/ssvc-calc/SSVC_JSON_2.0_min.schema.json 

Sample - https://github.com/CERTCC/SSVC/blob/main/ssvc-calc/sample-ssvc.json 

This was written mainly to make it easy for the tool to ingest a new decision tree.  The tool also exports part of the tree which was followed by an "analyst" which can also be represented in a JSON schema.  The JSON schema can also be built from the CSV input, but it will be missing some metadata on the decisions themselves.

j---:
I think you're right it should be pretty easy. The "guidance on communicating results" section (https://github.com/CERTCC/SSVC/blob/main/doc/version_1/047_treesForVulMgmt_4.md) is now mostly together, so we can make sure the schema and that section match, and then describe the schema a bit there. 
Loose requirements to check: 
 - [ ] Decision points are defined
 - [ ] Values for decision points are fixed
 - [ ] "Whole tree" schema is defined with some reasonable definition of when a tree is complete (all leaf nodes have a label, etc.)
 - [ ] Ideally, "whole tree" schemas can inherit  the definitions of decision points from one document, to make it easier to keep those definitions consistent
 - [ ] "Choice schema" for a specific set of choices of values for decision points and the resulting decision is defined and allows for the uncertainty in
 - [ ] "Choice schema" allows for identifying which "whole tree schema" was used 
 - [ ] Timestamps are required for the "choice" schema
 - [ ] "Choice" schema can handle partial information (see "Guidance..." section description)
 - [ ] Formatting (upper / lower case, etc.) match https://github.com/CERTCC/SSVC/blob/main/doc/style-guide-how-to (as much as makes sense, there is no *italics* in JSON)
 - [ ] Style guide documents that JSON literals are written in the .md files as `code literals`

sei-vsarvepalli:
I think it looks fine. 

Here is how I propose for next steps for which i will submit part of the PR.

[...]

PR #86, opened 2020-12-06
New JSON files for schema and examples

sei-vsarvepalli:
Request review and merge of this branch for updates to the /data folder - adding JSON schemas and examples of JSON files under folders "schema" and "computed".  These have been tested with JSON Schema tool for version 4.0 JSON schema. 

Note to remember and fix later: 
Even the latest JSON Schema (7) does not have a good way to represent Epoch in milliseconds and relating it to a timestamp.  So for now, Epoch timestamp remains an integer and the date field remains a normal date-time field - potentially using ISO 8601 time format.


j--- (approved):

j---:
I think this closes the substantive part of #65. However, to actually fully handle #65, we'll also need to edit the .md to talk about the JSON schema (which I plan to handle).

sei-vsarvepalli:
> I think this closes the substantive part of #65. However, to actually fully handle #65, we'll also need to edit the .md to talk about the JSON schema (which I plan to handle).

Hello Jono,

I have a simple README.md, it will be good to make it more exhaustive and clear about how it will be used.  The schema will probably also be iterative and may require improvements.  I will wait for @ahouseholder thoughts also.

Thanks
Vijay

data/schema/SSVC_Computed_v2.schema.json
```diff
@@ -0,0 +1,54 @@
...
+	    "description": "Computed score short representation such as SSVCv2/Ps:Nm/T:T/U:E/1605040000/ for a vulnerability with no or minor Public Safety Impact,  total Technical Impact, and efficient Utility, which was evaluated on Nov 10, 2020.",
+	    "type": "string"
+	},
+	"timestamp" : {
+	    "description": "Date and time in ISO format  ISO 8601 format",
+	    "type": "string",
+	    "format": "date-time"
+	},
+	"timestamp_epoch_ms" : {
+	    "description": "Date and time in epocj utc millisec",
```
ahouseholder, line 29: "epocj" should be "epoch"

sei-vsarvepalli: Thanks for this catch.  There is a little more consolidation of information I noticed in the Provision JSON schema.  I will update all this and push a new PR soon out.

ahouseholder (changes requested):
- [ ] _timestamp_ should be ISO 8601 with time zone to make it easier for receiver parsing
- [ ] _timestamp_epoch_ms_ appears redundant to _timestamp_ and is easy enough to get on the receiver end once _timestamp_ has been parsed. I'd prefer just having the one _timestamp_ field. 
- [ ] If characters in ISO8601 are a problem for the shorthand notation (which appears to append a timestamp to an ssvc string), I'd just make it positional: e.g., YYYYmmddHHMMSS. I don't see a need for subsecond resolution, and human-readability has value.

sei-vsarvepalli:
> * [ ]  _timestamp_ should be ISO 8601 with time zone to make it easier for receiver parsing
> * [ ]  _timestamp_epoch_ms_ appears redundant to _timestamp_ and is easy enough to get on the receiver end once _timestamp_ has been parsed. I'd prefer just having the one _timestamp_ field.
> * [ ]  If characters in ISO8601 are a problem for the shorthand notation (which appears to append a timestamp to an ssvc string), I'd just make it positional: e.g., YYYYmmddHHMMSS. I don't see a need for subsecond resolution, and human-readability has value.

Allen,
Thanks very much for the review and comments! 

One question I have is YYYYmmddHHMMSS should that be in UTC? 

I have a few more changes that I will consolidate into this PR (include your changes as well) this afternoon. Appending the timestamp epoch for short form of ssVC score  is mentioned in the below link. @j---  may need to update that document as well for ease of reading date-time in the shorthand score representation. 

https://github.com/CERTCC/SSVC/blob/main/doc/version_1/047_treesForVulMgmt_4.md

sei-vsarvepalli:
I am going to completely close this one and start a new PR, this branch is now all confusingly messed up.

Issue #79, opened 2020-11-24
Link mitigation to the SSVC decision points.


laurie-tyz:
- [x] For the defined mitigation actions, link them to the SSVC decision point that they can influence. 

(Associated with issue #46 )

j---:
This is related to #78. I'd say #79 is a prerequisite for #78. While I think #78 is not achievable in SSVCv2.0, this one does seem achievable. 

#79 has a dependency on moving the defined remediation actions to their own subsection within "Enumerating decisions". Although it puts things a bit out of order in the document, my thought at the moment is to address #79 in that new subsection. It's a bit out of order because "enumerating decisions" comes before the decision point definitions. But I think that is OK, and might be helpful, in that the reader will get a preview of the names of the decision points and what actions they are related to before we give the whole detailed description. This might be a helpful bit of signposting.

zmanion:
Link mitigation (not remediation) to decision points. E.g., a firewall rule changes/reduces System Exposure, then SSVC is assessed again, and a lower priority results.

laurie-tyz:
Wording I'm thinking of adding.


Deploying a mitigation reduces the system exposure. In the suggested tree the system exposure can be reduced from ‘open’ to ‘controlled’ or from ‘controlled’ to ‘small’.  Depending how your decision tree is constructed there may be a cascading effect on the other decisions. In the end, this reduction in System Exposure may or may not result in a change to the priority for further action.

from Jono:
Financial well-being impact might be reduced with adequate fraud detection and insurance. Physical well-being impact might be reduced by physical barriers that restrict a robot's ability to interact with humans. Mission impact might be reduced by introducing back-up business flows that do not use the vulnerable component.

laurie-tyz:
completed

PR #90, opened 2020-12-10
Related to item #79

laurie-tyz:
Some substitution of patch or fix for remediation or remediated.

one deletion in 047: deleted "applier".


j--- (changes requested):
I think we should try to get all the relevant instances of "patch" with this issue if it's part of 79. So in 040_treesForVulMgmt.md, the headings "supplying patches" etc. and the language in those sections needs to be updated to reflect we are talking about mitigations in general with SSVC. Currently those sections imply SSVC is only about patches.

laurie-tyz:
I don't believe these changes will fully close out the "patch" or "fix" issue.  I wanted to start slowly with *some* changes that appear obvious.  So a partial solution.  The rest of the plan was to start from the beginning (again) and identify those patch/fix instances that should really be mitigation/remediation.

Your call on how you want it done.

j--- (approved):
OK, if that's the plan then this looks good to me.

doc/version_1/047_treesForVulMgmt_4.md
```diff
@@ -71,7 +71,7 @@ Once the decision points are selected and the prioritization labels agreed upon,
 Making the decision process accessible has a lot of benefits.
 Unfortunately, it also makes it a bit too easy to overcomplicate the decision.
 
-The SSVC version 1 ~applier~ deployer tree had 225 rows when we wrote it out in long text form.
+The SSVC version 1 deployer tree had 225 rows when we wrote it out in long text form.
```
j---, line 74: I had this in strikeout (~tilde~) on purpose to make it clear that it was called something else in v1, so if a reader goes to look at the v1 doc they know it was called applier then, though we have changed it to deployer now. So I think this should stay. Sorry, I didn't catch this earlier.

Issue #88, opened 2020-12-10
Suggestions and feedback for "vector" formatting


ahouseholder:
Given the current state of the shorthand vector after #64, and in the process of reviewing #86 I have the following suggestions:

- [x] replace SSVCv2/... with either SSVC/2/... or SSVC/v2/... -- "ssvc" as a namespace, "version" is just a string
- [x] all timestamps should be in ISO 8601 YYYY-mm-ddTHH:MM:SS±TZ format including TZ, with UTC implied if no TZ specified
- (moved to #108) consider including a slot for a vul ID. This needn't be CVE, could be VU#, JVN#, or just a locally-relevant case number from a ticketing system or scan finding ID. I'd probably put it closer to the front than back, so maybe SSVC/v2/VU#99999/<vector>/<timestamp> 
- (moved to #109) The doc should specify whether order of parameters matters and how to determine that (is there a canonical ordering?) My own opinion is that order shouldn't matter.
- [x] If order doesn't matter then either timestamp needs to always be last or should be moved to the frontmatter of the string (not necessarily first position, but before the k:v starts) so that parsing code knows where key:value pairs stop
- [x] Are underspecfied vectors allowed? (e.g. if I only know two parameters can I still vectorize?) If so, how are gaps notated? (e.g., key:- means null value for key vs just omitting keys with null values from the vector)
- (moved to #110)  Even a single "version" of SSVC like v2 has multiple trees -- does the vector represent all of them simultaneously? Or do we need to be able to specify which tree within a version this applies to? (e.g., SSVC/v2/coordinate/vul_id/key:value.../timestamp)

Commentary:

- Separate namespace and version to simplify parsing
- We should avoid overspecifying numeric version strings, because the whole point of SSVC is customization and so there could be a lot of versions and we're not proposing to establish a versioning authority
- Make timestamp parsing easy by being really specific and consistent on what the format should be, and use a globally recognized standard that is guaranteed to have code support regardless of language or platform. Unix Epoch time seems inferior to TZ-aware ISO 8601 to me.
- Adding a vul ID breaks some format symmetry with CVSS vectors but I don't think there's a requirement to maintain that degree of symmetry in the first place. 
- People will come up with creative interpretations of any gaps in our spec, and that can inhibit interoperability in the long run.

ahouseholder:
With the separation of the remaining issues in #108, #109, and #110, I think this one is ready for closure by #107.

PR #91, opened 2020-12-10
Schemas updated and merged adh suggestions excep vector representatio…

sei-vsarvepalli:
sorry for the confusion from #86 issue.  There were too many changes and conflicts in one go.  All the suggestions are now in one place.  Only thing that remains unresolved in the #88 which will update the vector representation of the SSVC score for a "computed" vulnerability SSVC score.

Vijay


data/schema/SSVC_Provision_v2.schema.json
```diff
@@ -0,0 +1,119 @@
+{
+    "$schema": "http://json-schema.org/draft-04/schema#",
+    "$id": "https://democert.org/ssvc/SSVC_Provision_v2.schema.json",
+    "title": "Decision tree schema definition for SSVC v2",
+    "description": "This provides a schema for a decision tree used to compute SSVC score for a vulnerability",
+    "definitions": {
+        "color_regex": {
```
ahouseholder, line 7: What function does the color regex serve? This seems like it belongs in a UI, not in a data schema. Maybe I'm missing some context?

ahouseholder: I should also mention that I don't think this color regex was _added_ by this pull request, it appears to have been there previously. This is just where I noticed it. I don't have any problems with the intended changes otherwise, so if you'd prefer to merge this one and deal with the color regex question separately (e.g., in a separate issue), that's fine with me.

j--- (approved):
I think this resolves the issue in question. @ahouseholder color question should be a new issue.

data/schema/SSVC_Provision_v2.schema.json
sei-vsarvepalli: Allen,
The color regex is for UI - specifically useful in my UI provided to CISA https://democert.org/sssvc/   - surely.  It is an optional field and does not need to be there. It is convenient for the UI that I developed to communicate that's all.

I will leave it there and when we come back we can clean it up. If you think the stretch is much to include UI elements for full description of SSVC decision tree.

Vijay

PR #92, opened 2020-12-15
Item #79 part 2

laurie-tyz:
Focused only on 040_decisionTrees... section.

Limited changes from "patch" to "remediate"
Limited changes from "fix" to "remediate"
Limited changes from "fix" to "patch"

This is only for section 040.
Section 045 is free of references to patch or fix
Sections 047 & 050 will be reviewed next.

Thanks,
Laurie


cgyarbrough:
I think that the language in DoD Instruction 8531.01 is consistent with Laurie's changes, with the exception of the last one 'Limited changes from "fix" to "patch". The important differentiation between  mitigation and remediation is that mitigations attempt to reduce risk to a vulnerability that cannot be remediated (I read that remediate = fix). While patch is a well-known term, it is not directly defined in DoDI 8531.01. It relies on the definitions in CNSSI 4009 to define patch. According to the DODI 8531.01 language, patch may be considered a remediation if it closes a vulnerability, but it could also be part of a mitigation if the vulnerability was more complex and needed more than just a patch.

laurie-tyz:
Cancelling pull request.

Changes will be made to the next version to address @cgyarbrough comments.

PR #95, opened 2021-01-07
reorg, updated compile, updated README

j---:


ahouseholder (approved):
Looks fine to me, although I was anticipating a few changes to the contents of the .md files to as part of the reorganization. But if this PR is just about moving files around that's fine too, no objections from me.

PR #96, opened 2021-01-07
discuss changing information 

### LINKED ISSUE #29: Changing information
How are timestamps handled? How do users managing changing information? How is changed or changeable information documented?
sub-part of #24

j---:
fix #29

### DIFF
--- a/doc/md_src_files/060_decision-trees.md
+++ b/doc/md_src_files/060_decision-trees.md
@@ -205,7 +205,7 @@ The following decision points are usually in the control of the organization run
  - [*Mission Impact*](#mission-impact)
  - [*System Exposure*](#system-exposure)
 
-If SSVC information is all timestamped appropriately, then an analyst can compare the timestamp to the current date and determine whether information is considered stale.
+If SSVC information is all timestamped appropriately (as discussed earlier in this section), then an analyst can compare the timestamp to the current date and determine whether information is considered stale.
 The above rates are heuristic suggestions, and organizations may choose different ones.
 Any public repository of vulnerability information should keep a change log of when values change for each decision point, for each vulnerability.
 Vulnerability response analysts should keep such change logs as well.
--- a/doc/md_src_files/060_decision-trees.md
+++ b/doc/md_src_files/060_decision-trees.md
@@ -182,11 +182,18 @@ Both of these examples are out of the direct control of the vulnerability manage
 Some, such as [*Exposure*](#exposure), change mostly due to actions taken by the organization managing the vulnerable component.
 If the actor who can usually trigger a relevant change is the organization using SSVC, then it is relatively straightforward to know when to update the SSVC decision.
 That is, the organization should reevaluate the decision when they make a relevant change.
-For those decision points that are about topics outside the control of the organization using SSVC, then the organization will have to occasionally poll their information sources for changes.
+For those decision points that are about topics outside the control of the organization using SSVC, then the organization should occasionally poll their information sources for changes.
 The cadence or rate of polls is different for each decision point, based on the expected rate of change.
+
+We expect that updating information over time will be most useful where the evidence-gathering process can be automated.
+Organizations that have mature asset management systems will also find this update process more efficient than those that do not.
+For an organization without a mature asset management system, we would recommend putting organizational resources into maturing that function before putting effort into regular updates to vulnerability prioritization decision points.   
 
 The following decision points are usually out of the control of the organization running SSVC.
 As an initial heuristic, we suggest the associated polling frequency for each.
+These frequencies can be customized, as the update frequency is directly related to the organization's tolerance for the risk that the information is out of date.
+As discussed in [Tree Construction and Customization Guidance](#tree-construction-and-customization-guidance), risk tolerance is unique to each organization.
+Risk tolerance and risk appetite are primarily reflected in the priority labels (that is, decisions) encoded in the SSVC decision tree, but information polling frequency is also a risk tolerance decision and each organization may choose different time values. 
  - [*State of Exploitation*](#state-of-exploitation): every 1 day
  - [*Technical Impact*](#technical-impact): never (should be static per vulnerability)
  - [*Utility*](#utility): every 6 months
@@ -201,7 +208,8 @@ The following decision points are usually in the control of the organization run
 If SSVC information is all timestamped appropriately, then an analyst can compare the timestamp to the current date and determine whether information is considered stale.
 The above rates are heuristic suggestions, and organizations may choose different ones.
 Any public repository of vulnerability information should keep a change log of when values change for each decision point, for each vulnerability.
-Vulnerability response analysts may or may not need to keep such change logs, but it would facilitate after-action reviews if responders keep such change logs. 
+Vulnerability response analysts should keep such change logs as well.
+Similar to logging practices recommended for incident response [@nist800-61r2], such practices make the process less error-prone and facilitate after-action reviews.
 
 
 ## Development Methodology
--- a/doc/md_src_files/060_decision-trees.md
+++ b/doc/md_src_files/060_decision-trees.md
@@ -146,9 +146,9 @@ SSVCv2/Ps:Nm/T:T/U:E/1605040000/
 ```
 For a vulnerability with [no or minor](#public-safety-impact) [*Public Safety Impact*](#public-safety-impact), [total](#technical-impact) [*Technical Impact*](#technical-impact), and [efficient](#utility) [*Utility*](#utility), which was evaluated on Nov 10, 2020.
 
-    - TODO if we are going to talk about JSON or other structured data formats for decisions, do so here.
+    - TODO fix #65 (JSON format description) here.
 
-What an analyst knows about a vulnerability may not be complete. 
+What an analyst knows about a vulnerability may not be complete.
 However, the vulnerability management community may still benefit from partial information.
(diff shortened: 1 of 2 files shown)


doc/md_src_files/060_decision-trees.md
```diff
@@ -171,8 +171,38 @@ In abbreviated form, write this as `U:LE`.
...
+SSVC decision points have different temporal properties.
+Some, such as [*Utility*](#utility), are mostly static.
+For [*Utility*](#utility) to change, the market penetration or deployment norms of a vulnerable component would have to meaningfully change.
+Some, such as [*State of Exploitation*](#state-of-exploitation), may change quickly but only in one direction.
+
+Both of these examples are out of the direct control of the vulnerability manager.
+Some, such as [*Exposure*](#exposure), change mostly due to actions taken by the organization managing the vulnerable component.
+If the actor who can usually trigger a relevant change is the organization using SSVC, then it is relatively straightforward to know when to update the SSVC decision.
+That is, the organization should reevaluate the decision when they make a relevant change.
+For those decision points that are about topics outside the control of the organization using SSVC, then the organization will have to occasionally poll their information sources for changes.
```
cgyarbrough, line 185: I am concerned that the periodic reevaluation of decision points may introduce a resource burden that may be objectionable to users of SSVC. Perhaps the reevaluation should be labelled as a 'rescoring' and be described as a discrete decision process and not part of a periodic reevaluation. Additionally, SSVC scoring will vary even within an organization if their choice of information sources changes.

doc/md_src_files/060_decision-trees.md
```diff
@@ -171,8 +171,38 @@ In abbreviated form, write this as `U:LE`.
...
+ - [*Public Safety Impact*](#public-safety-impact): every 1 year
+
+The following decision points are usually in the control of the organization running SSVC and should be reevaluated when a relevant change is made or during annual reviews of assets.
+
+ - [*Situated Safety Impact*](#situated-safety-impact)
+ - [*Mission Impact*](#mission-impact)
+ - [*System Exposure*](#system-exposure)
+
+If SSVC information is all timestamped appropriately, then an analyst can compare the timestamp to the current date and determine whether information is considered stale.
+The above rates are heuristic suggestions, and organizations may choose different ones.
```
cgyarbrough, line 202: So the determination of whether a timestamp is stale or not is idiosyncratic to each organization? Might an SSVC evaluation for one organization counter one from another organization, based on differing definitions of 'stale'? This would make SSVC scores much less useful for sharing between organizations. The timestamps are crucial, but need to be implemented uniformly.

doc/md_src_files/060_decision-trees.md
```diff
@@ -171,8 +171,38 @@ In abbreviated form, write this as `U:LE`.
...
+Any public repository of vulnerability information should keep a change log of when values change for each decision point, for each vulnerability.
+Vulnerability response analysts may or may not need to keep such change logs, but it would facilitate after-action reviews if responders keep such change logs.
```
cgyarbrough, line 204: I would argue that Vulnerability response analysts should be required to keep change logs, just like incident response analysts are. Chain of custody and longitudinal tracking of the history of a vulnerability and its role in a specific vulnerability management process/chain would depend on keeping an accurate timestamp of activity.

doc/md_src_files/060_decision-trees.md
j---: These seem like two separate points worth discussion:
 - resource concerns around re-evaluation or rescoring. 
 - changing information sources. 

I think this section should be written to address the first, but the latter I think is out of scope. 
The frequencies I suggest for re-evaluation relate to our automation expectations. Exploitation can be run daily since that's a script (related to https://www.usenix.org/conference/cset20/presentation/householder). The expectation I have for the others is that an asset management system is helping connect systems to open vuls. So I'm essentially expecting a yearly re-evaluation of safety impact per asset per year. Is that asking too much? Does that framing need to be explicit in the section?

doc/md_src_files/060_decision-trees.md
j---: This sort of thing is why we should provide heuristics on "stale," as we do. I don't think more is needed. Quoting later in the doc:
> Such customization is a central feature of SSVC, as described in [Tree Construction and Customization Guidance](#tree-construction-and-customization-guidance).

When something is stale is just another part of risk tolerance and risk appetite. Those are the two things we expect to change among orgs. Is there something we've missed there? Because in general, it seems like there is a lot of use in harmonizing the definitions of what goes into the risk management evaluation, even though risk tolerance is idiosyncratic to the org. 

> but need to be implemented uniformly.
Is seconds past the epoch in UTC not adequately uniform?

doc/md_src_files/060_decision-trees.md
j---: I'm OK with being a bit stronger on our recommendation here (s/may or may not/should/). Can you provide a normative reference for the IR case, so we can link to it?

doc/md_src_files/060_decision-trees.md
cgyarbrough: If that is your expectation (that there will be a strong automated/automatable asset management system) is then that needs to be made explicit. Perhaps that is done in other areas of SSVC. My experience is that asset management is typically a weak area in many organizations. If we assume that one is in place it may limit the usefulness of SSVC to just those orgs that do have such a system. CVSS, for all its shortcomings, is applicable to a broad customer base because it sets a more or less arbitrary score then propagates that. I know we are trying to get away from the nonsensical score, but our assumptions need to be made explicit. If we are doing that then I'm ok with this.

doc/md_src_files/060_decision-trees.md
cgyarbrough: You are correct in that risk tolerance and appetite are expected to change. Once again, if we state our assumptions in other areas of SSVC then I have no objection. A more or less unstated assumption difference between SSVC and CVSS is that CVSS is assumed to be 'portable' between orgs and SSVC is not since it relies on customizable variables within each organization's operating environment. Is that correct?

doc/md_src_files/060_decision-trees.md
cgyarbrough: Sure. Look at BNIST Sp800-61r2 (Computer Security Incident Handling Guide), Section 3.2.5, Incident Documentation for this recommendation. https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-61r2.pdf This doc explicitly states that proper documentation, along with accurate timestamps, is critical. I would argue that vulnerability response can be viewed as an extension (or at least follows the same logic) that IR does.

j---: Thanks. I'll get this changed / added, then resolve this conversation.

doc/md_src_files/060_decision-trees.md
j---: I'll put a link to Tree Construction and Customization Guidance in the re-evaluation discussion, to make this more explicit. 
I think our target with SSVC is to define a set of variables (calling them decision points) that are more or less static. Then we suggest how they might be combined to reach a decision, but the orgs can customize how the variables are combined. When data becomes stale seems to me to be in the category of "how the variables are combined." Does that seem right?

doc/md_src_files/060_decision-trees.md
j---: We should nudge people towards asset management. But I agree it would be a bit naive to expect strong asset management in place as a prereq.

doc/md_src_files/060_decision-trees.md
cgyarbrough: Yes, I think that is correct. The customizable nature of the 'variables or decision points' will be both a 'selling point' and a point of confusion with folks who are currently using/misusing CVSS. Education will need to be provided showing how SSVC is superior due to its customizability.

doc/md_src_files/060_decision-trees.md
cgyarbrough: Thanks.

doc/md_src_files/060_decision-trees.md
```diff
@@ -171,8 +171,46 @@ In abbreviated form, write this as `U:LE`.
...
+ - [*Utility*](#utility): every 6 months
+ - [*Public Safety Impact*](#public-safety-impact): every 1 year
+
+The following decision points are usually in the control of the organization running SSVC and should be reevaluated when a relevant change is made or during annual reviews of assets.
+
+ - [*Situated Safety Impact*](#situated-safety-impact)
+ - [*Mission Impact*](#mission-impact)
+ - [*System Exposure*](#system-exposure)
+
+If SSVC information is all timestamped appropriately, then an analyst can compare the timestamp to the current date and determine whether information is considered stale.
```
madcatter24, line 208: Would it be appropriate here (or elsewhere) to recommend those that use a timestamp to use the same timezone? We recommend earlier on to keep the definitions of the decisions the same so that they can be shared externally, so it could be beneficial to keep the timestamps in the same timezone too?

madcatter24 (approved):
These changes look great to me. #65 is still outstanding for this.

doc/md_src_files/060_decision-trees.md
ahouseholder: I mentioned timestamps in #88:

>  all timestamps should be in ISO 8601 YYYY-mm-ddTHH:MM:SS±TZ format including TZ, with UTC implied if no TZ specified

Issue #60, opened 2020-11-04
Analyze Gini impurity and other DT metrics on revised trees 


ahouseholder:
This will need to be done after most of the revisions are made to the csv files that actually determine the trees. 

We should analyze and report on:

- [ ] Gini impurity
- [ ] Information gain

in the trees. It may be that these analyses hint at improvements to the tree, such as branch ordering or redundancies. I did this for the v1 trees, and although it didn't result in any major changes it was a useful check that we weren't missing something important.

ahouseholder:
Whatever we do should be repeatable. This probably means that code will be added to /src (either as a script or jupyter notebook) that can do whatever analysis we come up with for the CSV files in /data.

Still working to come up with what the useful analysis bits actually _are_ though. It's easy to load the CSV into sklearn and generate a decision tree from it and look at the stats on that, but how does that help us relate to the trees represented by the CSVs? That's less clear at the moment.

ahouseholder:
I think I'm narrowing in on the following:

Given a tree in which each decision point (feature) represents an ordinal categorical value (i.e., you can sort the options from low-to-high), a "good enough" tree should have the property that fitting a decision tree via the information gain metric will result in each decision point having feature importance > 0.

Given the choice between two candidate trees for the same stakeholder decision, prefer the one with a more level distribution across its feature importance (lower variance). This is an indication that each decision point is contributing equitably to the final decision.

j---:
> Given the choice between two candidate trees for the same stakeholder decision, prefer the one with a more level distribution across its feature importance (lower variance). This is an indication that each decision point is contributing equitably to the final decision.

Is this an "all other things equal" sort of preference? 
I'm fine with this as long as we also say that the tree expresses an organization's risk tolerance and policy, and there is value to a clearly expressed and usable policy which is going to be more important than getting all the features to have 1/N importance. 
Can you also cite some evidence that evenly-distributed feature importance is preferable for these kinds of operations-research type decision trees please? I think that would help users understand.

PR #98, opened 2021-01-13
added draft labels for new deployer tree

### LINKED ISSUE #23: downsize options for applier tree
225 is too many possible combinations for the Applier tree, have fewer. 
Directly a result of #21

j---:
May close #23 

This version appears to have many fewer "immediate"s than v1. I think that is partly due to compressing the higher mission and safety impact values into "very high". 
There is also some substantive change from v1, such as making sure that *small* *Exposure* causes an adequate prioritization reduction, which is feedback we received in a few diffuse ways on v1.

I'd like to reduce the number of options here. However, perhaps we can close #23 for now, and reopen it if #60 indicates this tree has branches that don't convey much information.


ahouseholder:
@j---  : Git makes it look like everything in this PR is new. Your comment implies it's not. So If I want to compare the PDF in this PR to what it was before, what PDF should I be bringing up in the second window to do a side-by-side?

j---:
> @j--- : Git makes it look like everything in this PR is new. Your comment implies it's not. So If I want to compare the PDF in this PR to what it was before, what PDF should I be bringing up in the second window to do a side-by-side?

It replaces the deployer tree:
https://github.com/CERTCC/SSVC/blob/main/doc/md_src_files/gfx/vul-management_v1-6-page3.pdf
https://github.com/CERTCC/SSVC/blob/main/doc/md_src_files/gfx/vul-management_v1-6-page4.pdf
https://github.com/CERTCC/SSVC/blob/main/doc/md_src_files/gfx/vul-management_v1-6-page5.pdf

Issue #78, opened 2020-11-24
What mitigations may lead to the most efficient risk reduction.


laurie-tyz:
- [ ] Where the appropriate decision is "mitigate" use the connection between types of mitigation actions and the SSVC decision points to recommend heuristics about what mitigations may lead to the most efficient risk reduction. 

(Associated with Issue #46 )

j---:
I think this will be pretty hard to do well, and I'm not sure we can accomplish it for v2.0. Maybe we can mention it in Future work (#38) and leave it for a future version?

ahouseholder:
For housekeeping purposes, I removed this issue from the SSVC v2 milestone for now. We can always add it back if we decide to address it in this round.

PR #101, opened 2021-01-29
Info sources 27

### LINKED ISSUE #24: Explain how to communicate and share SSVC information
This has several sub-parts:
- [x] #25 
- [x] #26 
- [ ] #27 
- [x] #28 
- [x] #29 

This issue is cross-referenced with #11 , because coordinators are one important stakeholder who may want to communicate SSVC info.

### LINKED ISSUE #27: Information sources for trees
For each decision point, explain a bit better how information to reach a value for that decision point can / should be gathered.
If people follow this guidance, what confidence can be shared when sharing SSVC information? 
sub-part of #24

j---:
fixes #27 
Because 27 is it's last part, closes #24

### DIFF
--- a/doc/md_src_files/050_decision-points_1.md
+++ b/doc/md_src_files/050_decision-points_1.md
@@ -17,6 +17,26 @@ The intent of this measure is the present state of exploitation of the vulnerabi
 | PoC <br /> (Proof of Concept) | One of the following cases is true: (1) exploit code sold or traded on underground or restricted fora; (2) typical public PoC in places such as Metasploit or ExploitDB; or (3) the vulnerability has a well-known method of exploitation. Some examples of condition (3) are open-source web proxies serve as the PoC code for how to exploit any vulnerability in the vein of improper validation of TLS certificates. As another example, Wireshark serves as a PoC for packet replay attacks on ethernet or WiFi networks. |
 | Active | Shared, observable, reliable evidence that the exploit is being used in the wild by real attackers; there is credible public reporting. |
 
+
+### Gathering Information About Exploitation
+[@householder2020historical] presents a method for searching the GitHub repositories of open-source exploit databases.
+This method could be leveraged to provide some information about whether [PoC](#exploitation) is true.
+However, part (3) of [PoC](#exploitation) would not be represented in such a search, so more information gathering would be needed.
+For part (3), perhaps we could construct a mapping of CWE-IDs which always represent vulnerabilities with well-known methods of exploitation.
+For example, CWE-295, [Improper Certificate Validation
+](https://cwe.mitre.org/data/definitions/295.html), and its child CWEs, describe improper validation of TLS certificates.
+These CWE-IDs could always be marked as [PoC](#exploitation) since that meets condition (3) in the definition.
+A comprehensive set of suggested CWE-IDs for this purpose is future work.
+
+Gathering information for [active](#exploitation) is a bit harder.
+If the vulnerability has a name or public identifier such as a CVE-ID, a search of news websites, twitter, the vendor's vulnerability description, and public vulnerability databases for mentions of exploitation is generally adequate.
+However, if the organization has the ability to detect exploitation attempts -- say through reliable and precise IDS signatures based on a public PoC -- then detection of exploitation attempts also signals that [active](#exploitation) is the right choice.
+
+The description for [none](#exploitation) says there is no **evidence** of exploitation.
+This framing admits that an analyst may not be able to detect or know about every attack.
+An analyst should feel comfortable selecting [none](#exploitation) if they (or their search scripts) have performed searches in the appropriate places for public PoCs and active exploitation (as described above) and found none.
+Acknowledging that [*Exploitation*](#exploitation) values can change relatively quickly, we recommend conducting these searches frequently -- if they can be automated to the organization's satisfaction, perhaps once a day (see also [Guidance on Communicating Results](#guidance-on-communicating-results)).  
+
 ## Technical Impact
 > Technical Impact of Exploiting the Vulnerability
 
@@ -35,7 +55,10 @@ Therefore, if there is a vulnerability then there must be some technical impact.
 | Partial | The exploit gives the adversary *limited* control over, or information exposure about, the behavior of the software that contains the vulnerability. Or the exploit gives the adversary an importantly low stochastic opportunity for total control. In this context, “low” means that the attacker cannot reasonably make enough attempts to overcome the low chance of each attempt not working. Denial of service is a form of limited control over the behavior of the vulnerable component. |
 | Total   | The exploit gives the adversary *total* control over the behavior of the software, or it gives total disclosure of all information on the system that contains the vulnerability       |
 
-## Utility 
+
+### Gathering Information About Technical Impact
+
+## Utility
 > The Usefulness of the Exploit to the Adversary
 
 [*Utility*](#utility) estimates an adversary's benefit compared to their effort based on the assumption that they can exploit the vulnerability.
@@ -49,13 +72,23 @@ Heuristically, we base [*Utility*](#utility) on a combination of value density o
 This framing makes it easier to analytically derive these categories from a description of the vulnerability and the affected component.
 [*Automatable*](#automatable) as ([*no*](#automatable) or [*yes*](#automatable)) and [*Value Density*](#value-density) as ([*diffuse*](#value-density) or [*concentrated*](#value-density)) are defined in Sections 4.4.3.1 and 4.4.3.2.
 
-Roughly, [*Utility*](#utility) is a combination of two things: (1) the value of each exploitation event and (2) the ease and speed with which the adversary can cause exploitation events. We define [*Utility*](#utility) as laborious, efficient, or super effective, as described in Table 6.
+Roughly, [*Utility*](#utility) is a combination of two things: (1) the value of each exploitation event and (2) the ease and speed with which the adversary can cause exploitation events. We define [*Utility*](#utility) as laborious, efficient, or super effective, as described in Table 6. Table 7 is an equivalent expression of [*Utility*](#utility) more like a lookup table a program might use.
 
 |  | Table 6: Utility Decision Values |
 | --------------- | ------------------------------------------------------------------------------ |
 | Laborious       | *No* to automatable and diffuse value                                               |
 | Efficient       | {*Yes* to automatable and diffuse value} OR {*No* to automatable and concentrated value} |
 | Super Effective | *Yes* to automatable and concentrated value                                         |
+
+Table 7: Utility to the Adversary, as a Combination of automatable and Value Density
+
+| *Automatable* | *Value Density* | *Utility* |
+| ----------- | --------------- |       --: |
+| *no*  | *diffuse*   | laborious |
+| *no*  | *concentrated* | efficient |
+| *yes* | *diffuse*   | efficient |
+| *yes* | *concentrated* | super effective |
+
 
 
 ### Automatable
@@ -76,11 +109,15 @@ Roughly, [*Utility*](#utility) is a combination of two things: (1) the value of 
 
 Due to vulnerability chaining, there is some nuance as to whether reconnaissance can be automated. For example, consider a vulnerability A. If the systems vulnerable to A are usually not openly connected to incoming traffic ([*Exposure*](#exposure) is [small](#exposure) or [controlled](#exposure)), reconnaissance probably cannot be automated (as scans should be blocked, etc.). This fact would make automatable [no](#automatable). However, if another vulnerability B with a [yes](#automatiability) to automatable can be reliably used to chain to vulnerability A, then that automates reconnaissance of vulnerable systems. In such a situation, the analyst should continue to analyze vulnerability A to understand whether the remaining steps in the kill chain can be automated.
 
+#### Gathering Information About Automatable
+
+
 Like all SSVC decision points, [*Automatable*](#automatable) should capture the analyst's best understanding of plausible scenarios at the time of the analysis.
 An answer of *no* does not mean that it is absolutely inconceivable to automate exploitation in any scenario.
 It means the analyst is not able to sketch a plausible path through all four kill chain steps.  
 Code that demonstrably automates all four kill chain steps certainly satisfies as a sketch.
 We say sketch to indicate that plausible arguments, such as convincing psuedocode of an automation pathway for each step, are also adequate evidence in favor of a *yes* to  [*Automatable*](#automatable).
+
 
 ### Value Density
 
@@ -109,17 +146,9 @@ We say sketch to indicate that plausible arguments, such as convincing psuedocod
     amount of data, but because it is uniquely valuable to law
     enforcement.
 
-The output for the [*Utility*](#utility) decision point is visualized in Table 7.
+#### Gathering Information About Value Density
 
-Table 7: Utility to the Adversary, as a Combination of automatable and Value Density
-
-| *automatable* | *Value Density* | *Utility* |
-| ----------- | --------------- |       --: |
-| *no*  | *diffuse*   | laborious |
-| *no*  | *concentrated* | efficient |
-| *yes* | *diffuse*   | efficient |
-| *yes* | *concentrated* | super effective |
-
+### Alternative Utility Outputs
 
 Alternative heuristics for proxying adversary utility are plausible. One such example is the value the vulnerability would have were it sold on the open market. Some firms, such as [Zerodium](https://zerodium.com/program.html), make such pricing structures public. The valuable exploits track the automatable and value density heuristics for the most part. Within a single system—whether it is Apache, Windows, iOS or WhatsApp—more automated kill chain steps successfully leads to higher exploit value. Remote code execution with sandbox escape and without user interaction are the most valuable exploits, and those features describe automation of the relevant kill chain steps. How equivalently virulent exploits for different systems are priced relative to each other is more idiosyncratic. Price does not only track value density of the system, but presumably also the existing supply of exploits and the installation distribution among the targets of Zerodium’s customers. Currently, we simplify the analysis and ignore these factors.
 However, future work should look for and prevent large mismatches between the outputs of the [*Utility*](#utility) decision point and the exploit markets.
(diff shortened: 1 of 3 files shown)


doc/md_src_files/050_decision-points_1.md
```diff
@@ -76,12 +109,16 @@ Roughly, [*Utility*](#utility) is a combination of two things: (1) the value of
 
 Due to vulnerability chaining, there is some nuance as to whether reconnaissance can be automated. For example, consider a vulnerability A. If the systems vulnerable to A are usually not openly connected to incoming traffic ([*Exposure*](#exposure) is [small](#exposure) or [controlled](#exposure)), reconnaissance probably cannot be automated (as scans should be blocked, etc.). This fact would make automatable [no](#automatable). However, if another vulnerability B with a [yes](#automatiability) to automatable can be reliably used to chain to vulnerability A, then that automates reconnaissance of vulnerable systems. In such a situation, the analyst should continue to analyze vulnerability A to understand whether the remaining steps in the kill chain can be automated.
```
cgyarbrough, line 122: One more aspect to consider, which builds on the previous sentence, is 'chain ability' of a given vulnerability. This is probably beyond the scope of this discussion, but certain vulnerabilities might be considered more 'beachhead' category (i.e., tending to allow for ingress to a system or software) and others are more effective at lateral movement or secondary instantiation in a blended or staged exploitation attempt. DoDCAR has done some initial work on the additive nature of exploits, but each of those is composed of underlying exploitable vulnerabilities.

doc/md_src_files/055_decision-points_2.md
```diff
@@ -306,3 +306,5 @@ Whether that mitigation allows the deployer to defer further action varies accor
 | Small       | Local service or program; highly controlled network       |
 | Controlled  | Networked service with some access restrictions or mitigations already in place (whether locally or on the network). A successful mitigation must reliably interrupt the adversary’s attack, which requires the attack is detectable both reliably and quickly enough to respond. *Controlled* covers the situation in which a vulnerability can be exploited through chaining it with other vulnerabilities. The assumption is that the number of steps in the attack path is relatively low; if the path is long enough that it is implausible for an adversary to reliably execute it, then *exposure* should be *small*. |
 | Open | Internet or another widely accessible network where access cannot plausibly be restricted or controlled (e.g., DNS servers, web servers, VOIP servers, email servers)  |
+
```
cgyarbrough, line 309: Have we given any consideration to the relative value of vulnerabilities for even the non-obvious stages of the Kill Chain? Reconnaissance stage is an example. Some Vulnerabilities may not be valuable from an attack standpoint, but very valuable in allowing for asset enumeration and secondary exploit targeting. Has anyone ever analyzed the monetary amount of early stage vulnerabilities vs mid-to-later stage (post delivery) ones?

doc/md_src_files/050_decision-points_1.md
j---: I've captured this under #102. It's a good idea, but will take some additional work, I think.

doc/md_src_files/055_decision-points_2.md
j---: I'm not familiar with anything specific. Bug bounty payouts (which we discuss as part of *value density*) give you some idea: https://zerodium.com/program.html . You can see the differential value between partial and complete-chain vuls in their value to zerodium. If you mean "value to the defender of closing a vul" in terms of most efficient improvement to defense posture, I think that is a question to capture for #78 .

doc/md_src_files/050_decision-points_1.md
cgyarbrough: Agreed. I think that there is a difference is scope when looking at using SSVC to measure a single vulnerability vs. using chained vulnerabilities in an exploit attempt. Since SSVC is looking to replace CVSS for single vulnerabilities, I think the scope is appropriate. Future work.

doc/md_src_files/050_decision-points_1.md
```diff
@@ -122,13 +122,18 @@ Due to vulnerability chaining, there is some nuance as to whether reconnaissance
...
-An answer of *no* does not mean that it is absolutely inconceivable to automate exploitation in any scenario.
-It means the analyst is not able to sketch a plausible path through all four kill chain steps.  
+An analyst should be able to sketch the automation scenario and how it either does or does not satisfy each of the four kill chain steps.
+Once one step is not satisfied, the analyst can stop and select [*no](#automatable).
 Code that demonstrably automates all four kill chain steps certainly satisfies as a sketch.
 We say sketch to indicate that plausible arguments, such as convincing psuedocode of an automation pathway for each step, are also adequate evidence in favor of a *yes* to  [*Automatable*](#automatable).
 
+Like all SSVC decision points, [*Automatable*](#automatable) should capture the analyst's best understanding of plausible scenarios at the time of the analysis.
+An answer of *no* does not mean that it is absolutely inconceivable to automate exploitation in any scenario.
+It means the analyst is not able to sketch a plausible path through all four kill chain steps.
```
cgyarbrough, line 132: I am sure we address this elsewhere in the discussion, but what are the four Kill Chain steps as we are defining them? Others will vary. Or are we just counting the post-delivery stages of the LM Kill Chain?

doc/md_src_files/050_decision-points_1.md
```diff
@@ -133,11 +133,11 @@ It means the analyst is not able to sketch a plausible path through all four kil
...
 Furthermore, just because two hosts are misconfigured to have expose a vulnerable service while 2 million are properly configured does not make discovery of the service automatable.
-Recall, as discussed in in [Reasoning Steps Forward](#reasoning-steps-forward), the analyst should consider *credible* effects based on *known* use cases of the software system in order to be pragmatic about scope and providing values to decision points. 
+Recall, as discussed in in [Reasoning Steps Forward](#reasoning-steps-forward), the analyst should consider *credible* effects based on *known* use cases of the software system in order to be pragmatic about scope and providing values to decision points.
 
 ### Value Density
 
-[*Value Density*](#value-density) is described as diffuse or concentrated:
+[*Value Density*](#value-density) is described as *diffuse* or *concentrated* based on the resources that the adversary will gain control over with a single exploitation event:
```
cgyarbrough, line 140: Are we defining a 'single exploitation event' as utilizing a single vulnerability or will chained vulnerabilities be considered (i.e., blended attack)?

doc/md_src_files/050_decision-points_1.md
j---: > steps 1-4 of the kill chain [@hutchins2011intelligence] for this vulnerability for some reason. These
    steps are reconnaissance, weaponization, delivery, and exploitation. 

Original definition, basically. First four (of 7) stages, so it is not post-delivery. Source: 
```
@article{hutchins2011intelligence,
  title={Intelligence-driven computer network defense informed by analysis of adversary campaigns and intrusion kill chains},
  author={Hutchins, Eric M and Cloppert, Michael J and Amin, Rohan M},
  journal={Leading Issues in Information Warfare \& Security Research},
  volume={1},
  pages={80},
  year={2011}
}
```
The last three stages, IIRC, are installation, command and control, and action on objectives.

doc/md_src_files/055_decision-points_2.md
cgyarbrough: Perhaps so. In reality though, bug bounties usually look for high value vulnerabilities which generally involve targeted software or systems. I think that vendors who pay bug bounties may not use the same logic as defenders of deployed systems when they allocate higher $$ to vulnerabilities payouts. Defenders usually value the 'beachhead' vulnerabilities which, if patched, disrupt the kill chain.

doc/md_src_files/050_decision-points_1.md
j---: I hadn't considered this question. But now that I'm prompted, I would say a blended attack counts as "one exploitation event". Does that cause any problems?

doc/md_src_files/055_decision-points_2.md
j---: OK, then I think this is something to capture for #78 but in regard to across multiple vuls, what will be most efficient (e.g., fixing beachheads) rather than just within a single vul, what is the most efficient way to tackle it (which I think is what #78 focuses on now).

doc/md_src_files/050_decision-points_1.md
cgyarbrough: It really depends on what you are trying to capture/measure. If effectiveness and morbidity of an exploitation event vs. the discrete value of a specific vulnerability is what you are wanting to measure then the exploit attempt would be appropriate. In reality severity of a vulnerability is linked to the exploit in which it is used. If you want to measure/capture the future damage a specific vulnerability (increased risk) then we might measure something very different--especially since a single vulnerability might be used in multiple exploitation events which may have very different outcomes.

j---: > have very different outcomes.

I know we're always dancing with this. Context matters. The same software can be used for very different purposes. But I think we've covered that where we say: 

> ## Reasoning Steps Forward
> 
> This aspect of scope is about immediacy, prevalence, and causal importance. Immediacy is about how soon after the decision point adverse effects should occur to be considered relevant. Prevalence is about how common adverse effects should be to be considered relevant. Causal importance is about how much an exploitation of the software in the cyber-physical system contributes to adverse effects to be considered relevant. Our recommendation is to walk a pragmatic middle path on all three aspects. Effects are not relevant if they are merely possible, too infrequent, far distant, or unchanged by the vulnerability. But effects are relevant long before they are absolutely certain, ubiquitous, or occurring presently. Overall, we summarize this aspect of scope as *consider credible effects based on known use cases of the software system as a part of cyber-physical systems*.

> It really depends on what you are trying to capture/measure.

Utility of the vul to the attacker in the context of the value density of instances of the vulnerable component =)

doc/md_src_files/050_decision-points_1.md
```diff
@@ -17,10 +17,30 @@ The intent of this measure is the present state of exploitation of the vulnerabi
...
+[@householder2020historical] presents a method for searching the GitHub repositories of open-source exploit databases.
+This method could be leveraged to provide some information about whether [PoC](#exploitation) is true.
+However, part (3) of [PoC](#exploitation) would not be represented in such a search, so more information gathering would be needed.
+For part (3), perhaps we could construct a mapping of CWE-IDs which always represent vulnerabilities with well-known methods of exploitation.
+For example, CWE-295, [Improper Certificate Validation
+](https://cwe.mitre.org/data/definitions/295.html), and its child CWEs, describe improper validation of TLS certificates.
+These CWE-IDs could always be marked as [PoC](#exploitation) since that meets condition (3) in the definition.
+A comprehensive set of suggested CWE-IDs for this purpose is future work.
+
+Gathering information for [active](#exploitation) is a bit harder.
```
ahouseholder, line 31: I don't disagree with any of the points made here. 

Do you think it's worth saying anything about the idea that direct detection of exploitation is hard and that a lot of the "evidence of exploitation" comes from a path of incident -> malware -> re -> which vuls are involved, which really implies that the exploit has been found in deployed malware. I'm not close enough to 2020 era incident forensics to say more about the details there. But I suppose the main gist is that the path between incidents you know about and knowing which vul(s) is(were) involved is often quite lossy -- and that often the best you can do is rely on publicly available reporting.

doc/md_src_files/050_decision-points_1.md
```diff
@@ -35,7 +55,21 @@ Therefore, if there is a vulnerability then there must be some technical impact.
 | Partial | The exploit gives the adversary *limited* control over, or information exposure about, the behavior of the software that contains the vulnerability. Or the exploit gives the adversary an importantly low stochastic opportunity for total control. In this context, “low” means that the attacker cannot reasonably make enough attempts to overcome the low chance of each attempt not working. Denial of service is a form of limited control over the behavior of the vulnerable component. |
 | Total   | The exploit gives the adversary *total* control over the behavior of the software, or it gives total disclosure of all information on the system that contains the vulnerability       |
 
-## Utility 
+
+### Gathering Information About Technical Impact
+
+Assessing [*Technical Impact*](#technical-impact) amounts to assessing the degree of control the vulnerability gives an attacker over the vulnerable component.
```
ahouseholder, line 61: "the degree of control over the vulnerable component the attacker stands to gain by exploiting the vulnerability"

vulnerabilities are options to control. exploits exercise the option.

doc/md_src_files/050_decision-points_1.md
```diff
@@ -35,7 +55,21 @@ Therefore, if there is a vulnerability then there must be some technical impact.
...
+One way to approach this analyiss is to ask whether the control gained is *total* or not.
+If it is not total, it is *partial*.
+If an answer to one of the following questions is yes, then control is *total*:
```
ahouseholder, line 64: if you accept the above comment's premise, then the bullets below might need to be prefixed with "after exploiting the vul, does the attacker gain..." or something similar. Point being to highlight the specific change in ability the attacker gains via exploitation (contrast with a possible reading that they have the ability with or without the vulnerability).

doc/md_src_files/050_decision-points_1.md
ahouseholder: Should we note that weaponization  might not be strictly necessary as a positive act? I.e., "wormable" vuls turn weaponization into a development process that happens once and then exploitation of vulnerable systems is reliable from that point on. A worm seems to automate recon, delivery, and exploitation, but I'm not sure that automation of the weaponization step was necessary. 

I may also be missing nuance in my understanding of how folks use the kill chain model.

doc/md_src_files/055_decision-points_2.md
```diff
@@ -306,3 +306,21 @@ Whether that mitigation allows the deployer to defer further action varies accor
...
+[*System Exposure*](#system-exposure) can be readily informed by network scanning techniques.
+For example, if the vulnerable component is visible on [Shodan](www.shodan.io) or by some other external scanning service, then it is [*open*](#system-exposure).
+Network policy or diagrams are also useful information sources, especially for services intentionally open to the Internet such as public web servers.
+An analyst should also choose [*open*](#system-exposure) for a phone or PC that connects to the web or email without the usual protections (for example, IP and URL blocking, updated firewalls, etc.).
+
+Distinguishing between [*small*](#system-exposure) and [*controlled*](#system-exposure) is more nuanced.
+If [*open*](#system-exposure) has been ruled out, some suggested heuristics for differentiating the other two are as follows.
+ - If the system's networking and communication interfaces have been physically removed or disabled, choose [*small*](#system-exposure).
+ - If [*Automatable*](#automatable) is [*yes*](#automatable), then choose [*controlled*](#system-exposure).
```
ahouseholder, line 323: I'm not sure I see the logic here. How does automatable=yes lead to controlled vs small?

ahouseholder (changes requested):
I had a few suggestions as inline comments in the changes. They're open for debate though so I'd be ok if they're given consideration and declined.

doc/md_src_files/050_decision-points_1.md
j---: If something happens once and then can be copied repeatedly without human intervention, I think that falls under the usual definition of automation. I don't know what "positive act" means there. It was done at some point during the exploit development process. Is that not an act some how? If it can just be copied for each new target, then it's automated. Things like ASLR might interfere with automation here, but again overcoming something like ASLR can be automated. 

If you want to suggest additional text for Automation, can you open a new issue?

doc/md_src_files/050_decision-points_1.md
j---: added to the end of the paragraph starting line 31.

doc/md_src_files/050_decision-points_1.md
j---: Added prefix of question before all bullets, for better parallelism.

doc/md_src_files/055_decision-points_2.md
j---: Because if automatable is *yes*, then reconnaissance is yes. And if the system can be automatically reconned, then I would say it's exposure is higher than small. 

Does that make sense? Do we need to write that in to the bullet?

ahouseholder: So I think you're saying that automatable is scoped to your environment vs. scoped to the vul? That is not what I expected. (I.e., "can be reconned _in your environment_" vs "can be reconned _in general, regardless whether you have mitigating configurations impeding that in your environment_")

ahouseholder: To answer your question directly though, I think whatever the reasoning is should be explicitly in the bullet.

ahouseholder:
commits in a7418370d2b8255aaa0d1f752f269452429ef8b3 look good to me

doc/md_src_files/055_decision-points_2.md
j---: Automatable is about the usual deployment case. So it is not tied to the specific deployment environment. 
If you have suggestions for situations to "guard" this scenario, to state heuristics that can push users to *small*, please add them. I clarified these should be considered in order.

j---: How's https://github.com/CERTCC/SSVC/pull/101/commits/64dae383cb22b98811003f83595efc22ccd8ed29 ?

doc/md_src_files/050_decision-points_1.md
ahouseholder: I think my confusion arises in that "weaponization" is a verb, but an exploit is an artifact. So automated exploitation can use the resulting artifact even if the production of the artifact (which I'm taking to be what "weaponization" means) itself was not automated.

The automation isn't necessarily doing anything to customize the exploit for the target, so it doesn't feel like weaponization is automated so much as it's just taken out of the process that needs to be automated. But yes, I acknowledge this concern is off-topic for _this_ PR.

doc/md_src_files/055_decision-points_2.md
ahouseholder: Yeah, that seems to capture it.

ahouseholder (approved):

ahouseholder:
I think all my concerns in this one are addressed. If @cgyarbrough confirms his approval this one should be ok to merge.

doc/md_src_files/050_decision-points_1.md
cgyarbrough: I approve this pull. Seems like for the discussions in scope that I'm good.

cgyarbrough:
I approve this pull.

ahouseholder:
squashed and merged.

Issue #106, opened 2021-02-16
Updates for JSON format


j---:
Summarizing a conversation between @sei-vsarvepalli and myself earlier today:

1. [x] consider choices as a dict of lists rather than list of single-element dicts; e.g., '''choices = { "exploitation": ["none","poc","active], ...etc '''
2. [x] consider changing the name of the object "choices" in the decision spec to "options" to avoid conflict with the name "choices" in the answers spec
3. [x] allow for out-of-order answers to questions within trees: While it's ok for the order of questions to be consistent in the JSON format that doesn't mean that answers will always answer q1 then q2 then q3 etc. 
4. [x] allow for underspecified answers: related to above, it should be possible to say things like "exploitation" = "none" or "poc" but not "active". If each value is a list of values, then this just means the cardinality of the list can be >1 up to the max of all the possible options for that item. Furthermore this just means that a fully-specified answer is a list with cardinality = 1.
5. (moved to #109)  in the vector notation, address underspecification as well: E.g., E:[N,P] for the previous example.
6. (moved to #109) in the vector notation, consider a wildcard character in place of a full list when a decision is entirely unspecified e.g., X:* instead of X:[A,B,C,D] if X has four options, but X:[A,B] if we know X is neither C nor D.
7. [x] add "key" to the full tree specification to decisions and options to their corresponding short-form key in vector notation. The goal is to allow full parsing of the vector notation to a verbose interpretation given the tree description.
8. (moved to #111)  We should allow for vector keys and possibly value keys to be more than 1 character. We can declare by convention a preference for single character keys, but being able to accomodate longer ones without having to rev the spec will likely be convenient in the future. We are likely to run out of letters in the decision options first because we're already using a sizable fraction of 26. Within any given decision it's really only necessary that the option values are unique to that decision (i.e., "E:A" is not the same "A" as "D:A", but we can't have two "D:A"s). And it seems like if we have a single decision with >26 options maybe we're doing something wrong.
9. (moved to #110) vector notation should specify both version and role so it's clear which tree the vector is addressing. Suggest sticking with "/" as field separator, and "k:v" as key:value syntax for these. E.g., /V:2/R:C/
10. (moved to #110) in the ensuing discussion of the above, we realized we might have a need for tree IDs in the future, to avoid having to rev SSVC just because we change an option in the Coordinator tree for example. That's likely a future request, but I'm noting it here first just to get it written down.
[...]

PR #107, opened 2021-02-16
Proposed fix for 106 json format

sei-vsarvepalli:
This should address a bulk of the issues that were discussed with @ahouseholder in #106  . I will keep marking the ones that are addressed and capture them in the follow-up comments.


data/examples/Provision-Coordinator-v2.01.json
```diff
@@ -3,88 +3,107 @@
...
 		{
 		    "label": "active",
+		    "key": "A",
 		    "description": "Shared, observable, reliable evidence that the exploit is being used in the wild by real attackers; there is credible public reporting."
 		}
 	    ]
 	},
 	{
 	    "label": "Virulence",
+	    "key": "V",
```
j---, line 27: In SSVC v2 "Virulence" is being replaced by "Automatable" with options *yes* or *no*. I think that's out of scope for this issue, though.

j--- (approved):
I think this looks good. 
@sei-vsarvepalli , I'd like to talk about opening a new issue to make sure we document the way that the schema can be used with the CSV to make a "computed" entry that abides the correct style and forces entry elements to match the available decision point options. Another option would be to produce schema similar to examples/Provision-Coordinator for the recommended Deployer and Supplier trees in the v2 doc.

j---:
@ahouseholder , can you deconflict the open (unchecked) items in #106  with #88, and get us one list of what we need to address? It's not clear to me whether this PR should close 106, because everyone open is covered under #88, or not.

data/examples/Provision-Coordinator-v2.01.json
sei-vsarvepalli: There are a few ways to make the CSV also have a computed form. One of them is a add a boolean column called "active" or "available" with yes/no.  In CSV, each row represent a possible path in the tree and if the "available" column is "no" this option has been negated or disabled.  I think this may get a bit messy to organize in CSV. 

The other way is ...
We can generate CSV from JSON and have it eliminate the irrelevant/unavailable rows OR mark them such  (like the above added column).   The computed JSON with full provision tree can be the "system of record" for a vulnerability at a time, capturing more information about all possible paths in the tree and the ones that were "triaged" in some way.

I am open to both.  My only concern is the analyst ability to "create" computed JSON's or CSV's. It is going to be difficult without a tool to do this.

ahouseholder:
With the separate creation of #108, #109, and #110, I think this might fix #88.

data/examples/Provision-Coordinator-v2.01.json
ahouseholder: Suggestion: just create the issue and move these comments there. (Well, that probably means copy them and then do a _resolve conversation_ here.)

data/examples/Computed+Provision-VU#290915_Coordinator-v2.01.json
```diff
@@ -0,0 +1,396 @@
+{
+  "timestamp": "2020-12-10T18:58:04.153Z",
+  "role": "Coordinator",
+  "id": "VU#290915",
+  "version": "2.0",
+  "computed": "SSVCv2/E:P/V:R/T:P/M:H/D:A/1607626684/",
```
ahouseholder, line 6: I'd still like to see version be separate from "SSVC", which I take as a namespace rather than a version, but that's also covered in #110 so I'm ok with making that be a separate task if the decision is to defer it here.

ahouseholder: Similarly, indication of which tree the vector addresses, also in #110. A separate pull request later for that is fine too.

data/examples/Computed+Provision-VU#290915_Coordinator-v2.01.json
```diff
@@ -0,0 +1,396 @@
...
+      "Mission &amp; Well-being": "high"
+    },
+    {
+      "Decision": "Attend"
+    }
+  ],
+    "decision_tree": {
+    "decision_points": [
+	{
+	    "label": "Exploitation",
```
ahouseholder, line 27: Should individual decision points also have "descriptions" the way their options do?

data/schema/SSVC_Computed_v2.01.schema.json
```diff
@@ -9,15 +9,20 @@
...
+			}
+		    }
 		}
 	    },
 	    "minItems": 1,
 	    "uniqueItems": true		
 	},
 	"computed": {
-	    "description": "Computed score short representation such as SSVCv2/Ps:Nm/T:T/U:E/1605040000/ for a vulnerability with no or minor Public Safety Impact,  total Technical Impact, and efficient Utility, which was evaluated on Nov 10, 2020.",
+	    "description": "Computed score short representation such as SSVC/v:2/R:C/E:[P,A]/V:R/T:P/M:H/D:[T,R,A,E]/1607626684/ for a vulnerability with no or minor Public Safety Impact,  total Technical Impact, and efficient Utility, which was evaluated on Nov 10, 2020.",
```
ahouseholder, line 25: looks like you have version and role split out here, but not in https://github.com/CERTCC/SSVC/pull/107/files#diff-59989ecd9b70fdbb329640c6da907e5fcc247ba37bc20eaeac267f325e7f70e0 so this might just be a cleanup for consistency before merging this PR.

ahouseholder (changes requested):
There's an inconsistency in vector notation -- SSVC/v:2/R:C/<other k:v>... appears in one place but SSVCv2/<other k:v>... in others. I think this is just a change toward the former that wasn't propagated to the other json files.

Once that's fixed, then this PR will also fix #110 when merged.

data/examples/Computed+Provision-VU#290915_Coordinator-v2.01.json
ahouseholder: (If so, please cut a new issue for that so we can merge this and move on.)

data/examples/Computed+Provision-VU#290915_Coordinator-v2.01.json
j---: I think SSVC/2/... makes sense, that is, i'm in favor of separating the name space and the version.

data/schema/SSVC_Computed_v2.01.schema.json
sei-vsarvepalli: All changes look right.  One thing I want to be clear is v 2.01 mentioned is about the JSON file, it is NOT a new version of SSVC.  Sorry I should picked a different number scheme perhaps. Hope that makes sense.

It will be worth in next two weeks or so for three of us to meet and sort of clarify what our thoughts about things to cover and close in the next few weeks.

Issue #38, opened 2020-09-29
Rewrite the future work and limitations sections


j---:
This is a one issue because future work and limitations sections are closely related. Future work is problems we can fix. Limitations is problems we know we can't fix because we have made a choice to prioritize something else. 

Both of these sections will need to be updated to make sure that all items identified as future work in the rest of the document are collated. And in general, the v2 updates will change what needs to be done in the future and may adjust the limitations. In that sense, a final pass on these sections needs to be one of the last things done in v2.

- [x] Collate items identified as future work in the document to make sure nothing flagged as future work in prior sections is left out of the future work section. The main section headings in the Future Work section should still be used to emphasize the most important items. 
- [x] After most v2 work is done, edit / rewrite the future work section
- [x] After most v2 work is done, edit / rewrite the limitations section

laurie-tyz:
List of "Future Work" 

- [ ] VRDA is likely adequate for national-level CSIRTs that do general CVD, but other CSIRT types may have different needs. Future work may elicit those types and make a few different decision options.

- [ ] Specialized coordination organizations exist (e.g., ICS-CERT, which conducts CVD for safety-critical systems). We have not developed a coordination tree in this work, but future work could use our principles and design techniques to refine and evaluate VRDA or some other decision tree for coordinated vulnerability disclosure.

- [ ] Different risk tolerance or risk discounting postures are not addressed in the current work; accommodating such tolerance or discounting explicitly is an area for future work.

- [ ] However, future work should look for and prevent large mismatches between the outputs of the utility decision point and the exploit markets.

- [ ] Aggregation suggests that the stakeholder’s response to this decision point cannot be less than the most severe credible safety impact, but we leave the specific aggregation method or function as a domain-specific extension for future work.

- [ ] The factors that influence the safety impact level are diverse. This paper does not exhaustively discuss how a stakeholder should answer a question; that is a topic for future work.

- [ ] The factors that influence the mission impact level are diverse. This paper does not exhaustively discuss how a stakeholder should answer a question; that is a topic for future work.

- [ ] This paper focuses only on the structure of decision making. Advice about how to search for information about a vulnerability is a separate project that will be part of future work. In some domains, namely exploit availability, we have started that work in parallel.

- [ ] Future work on SSVC as well as further development of other prioritization methods can benefit from using the method described in the pilot. Future instances should use more representative participant demographics.

[...]

PR #119, opened 2021-03-03
update abbreviated format description

### LINKED ISSUE #109: Vector docs should specify whether parameter order and inclusion is required
Breaking this out from #88 to make it separately decidable

The docs should specify whether order of parameters matters and how to determine that (is there a canonical ordering?) My own opinion is that order shouldn't matter.

Closely related is whether all keys are required for the vector or if some may be omitted when information is only partially specified.

Questions that should answer:

- Must vectors always keep key:value pairs in the same order?
- How to handle unspecified elements? This is the omit/include-but-indicate-unknown item. E.g., what if I don't know the answer to Exploitation but I have answers to the others?
- How to handle underspecified elements? E.g., what if I know Exploitation is not PoC but I don't know if it's none or active? Can I use the vector indicate what I do know?

### LINKED ISSUE #111: Vector keys should allow more than 1 character
Breaking this out of #106 for separate decidability

We should allow for vector keys and possibly value keys to be more than 1 character. We can declare by convention a preference for single character keys, but being able to accommodate longer ones without having to rev the spec will likely be convenient in the future. 

We are likely to run out of letters in the decision options first because we're already using a sizable fraction of 26 across all the trees. 

Within any given decision it's really only necessary that the option values are unique to that decision (i.e., "E:A" is not the same "A" as "D:A", but we can't have two "D:A"s). And it seems like if we have a single decision with >26 options maybe we're doing something wrong.

j---:
should close #109  and close #111


sei-vsarvepalli:
Hey @j--- 
Looks good mostly.  Only question I have is on the language issue.  JSON format allows for UTF-8 representation of each of "keys" and "values"  in both the Provision and Computed schema.  So, what should happen if language is say Japanese in UTF-8, do they pick equivalent English letters?  Is SSVC vector only is limited to POSIX locale - basically US-ASCII chars only?  Maybe a short note of this would be good to clarify how short notation handles languages.

Vijay

ahouseholder (approved):
looks fine to me

ahouseholder:
> Looks good mostly. Only question I have is on the language issue. JSON format allows for UTF-8 representation of each of "keys" and "values" in both the Provision and Computed schema. So, what should happen if language is say Japanese in UTF-8, do they pick equivalent English letters? Is SSVC vector only is limited to POSIX locale - basically US-ASCII chars only? Maybe a short note of this would be good to clarify how short notation handles languages.

While I agree that i18n of the vector is an issue we should address eventually, it seems like a gap we might note in future work (i.e., #38) for now and then revisit it once we have some engagement from folks who need it. I'm in favor of cutting a new issue for internationalization of SSVC and appending this as one of the items for that issue to address.

sei-vsarvepalli:
> > Looks good mostly. Only question I have is on the language issue. JSON format allows for UTF-8 representation of each of "keys" and "values" in both the Provision and Computed schema. So, what should happen if language is say Japanese in UTF-8, do they pick equivalent English letters? Is SSVC vector only is limited to POSIX locale - basically US-ASCII chars only? Maybe a short note of this would be good to clarify how short notation handles languages.
> 
> While I agree that i18n of the vector is an issue we should address eventually, it seems like a gap we might note in future work (i.e., #38) for now and then revisit it once we have some engagement from folks who need it. I'm in favor of cutting a new issue for internationalization of SSVC and appending this as one of the items for that issue to address.

Cool - we can track it in the "internationalization" bucket.

sei-vsarvepalli (approved):
Internationalization will be addresses in an independent issue.

Issue #122, opened 2021-03-05
Summarize changes from v1 in v2 doc


j---:
Essentially summarize 
https://github.com/CERTCC/SSVC/issues?q=is%3Aissue+is%3Aclosed+project%3ACERTCC%2FSSVC%2F1

This is alluded to in #34 , but it should be a new issue.

PR #124, opened 2021-03-08
update the introduction with basic fixes.

### LINKED ISSUE #34: Re-write the introduction
Version 2 will need different context, different signposting, and in general a different introduction than v1. 
Some items to address include: 
 - [x] v1 and v1.1 have different kinds of intros and context. v1 has more decision tree basics. v1.1 has more background on operational and academic related work. Decide which of these to keep; maybe merge and keep both?
 - [x] Add the following to related work: Wang, B., Li, X., de Aguiar, L. P., Menasche, D. S., & Shafiq, Z. (2017). Characterizing and modeling patching practices of industrial control systems. Proceedings of the ACM on Measurement and Analysis of Computing Systems, 1(1), 1-23.
 - [x] Focus less on CVSS, any discussion of it moves to #30 
 - [x]  reflect new section structure (from #95)
 - Focus on what SSVC v2 provides. Do not focus on what SSVC v1 did, or what changes were made. Those go in a separate section, towards the end of the document handled by #122.

j---:
fixes #34 

This doesn't do anything to handle a change log or point to it. In fact, the text doesn't acknowledge that this is v2. I'm open as to whether it should, but it seemed like the best thing to do was to write the intro to the doc as it is, and handle introducing the history and change log in response to #122.


cgyarbrough:
I think that this looks good to me as well.

Issue #131, opened 2021-03-17
Organize / clean up tree CSV files


ahouseholder:
This arises from working on #60 but it's a separable task.

- Put v1 and v2 CSVs into separate directories within /data
- Remove any extra or obsolete csv files (Input from @j--- might be necessary here because the filenames are not always clear as to the files' purpose) 
- normalize column header capitalization scheme (pick one and apply it consistently)
- normalize column names where possible (e.g., is it _decision_ or _outcome_? Pick one and stick with it.)
- modify code to look in the new locations where necessary
- decide whether to keep "row" column (does the csv need a column for row numbers?)
- add a README.md file to the data/v2 dir with a brief explanation of what each file is for

PR #134, opened 2021-03-19
Reorganize CSV files

ahouseholder:
Fix for #131 

- [x] Put v1 and v2 CSVs into separate directories within /data
- [x] Remove any extra or obsolete csv files (Input from @j--- might be necessary here because the filenames are not always clear as to the files' purpose) 
- [x] normalize column header capitalization scheme (pick one and apply it consistently)
- [x] normalize column names where possible (e.g., is it _decision_ or _outcome_? Pick one and stick with it.)
- [x] modify code to look in the new locations where necessary
- [x] decide whether to keep "row" column (does the csv need a column for row numbers?)
- [x] add a README.md file to the data/v2 dir with a brief explanation of what each file is for


j---:
I think ssvc_2_deployer.csv and ssvc_2_deployer_simplified.csv are out of date. 
The wonky name for the other deployer tree is an attempt to use the initialisms of the decision points in the tree in the name, but that may be more confusing than it's worth. 

I think deployer-options_v2.csv is essentially a blank, it has no priority labels. That may not be necessary, but it might be helpful for people to customize starting from that. Not sure. 

The .tex file should be redundant, as it should be recreatable with src/SSVC_csv-to-latex.py.
The Coordinator triage tree has some custom latex to make it work, but for those situations I think that custom .tex should go in doc/graphics/ anyway.

ahouseholder:
So here's my proposed file cleanup, it'll be in the next commit.

Obsolete/Remove
* deployer-options_v2.csv 
* ssvc_2_deployer.csv - Deployer Priority, includes separate mission and safety impact subcomponents
* ssvc_2_deployer_simplified.csv - Deployer Priority, only consolidated safety/mission impact
* ssvc_2_deployer_SeUMss.tex

Keep as Is
* ssvc_2_coord-publish.csv - Coordinator Publication Priority 
* ssvc_2_coord-triage.csv - Coordinator Case Priority
* ssvc_2_supplier.csv - Supplier Priority, with all subcomponents
* ssvc_2_supplier_simplified.csv - Supplier Priority with just Exploitation,Utility,TechnicalImpact,PublicSafetyImpact

Keep but rename
* ssvc_2_deployer_SeEUMss.csv - Deployer Priority, with attacker utility + well-being/mission impact

ahouseholder:
Should we also eliminate ssvc_2_supplier.csv and keep only ssvc_2_supplier_simplified.csv? If so, I'll rename the simplified one to just be ssvc_2_supplier.csv

ahouseholder:
I think I've addressed the file moves in the python in /src. 
- ssvc.py needed to be pointed to the new location.
- SSVC_csv-to-latex.py doesn't appear to need any changes because the file is a command line argument.

Question for @sei-vsarvepalli - Do these changes affect anything in your javascript in ssvc-calc? I looked for anywhere the files were loaded but couldn't find anything.

j---:
> Should we also eliminate ssvc_2_supplier.csv and keep only ssvc_2_supplier_simplified.csv? If so, I'll rename the simplified one to just be ssvc_2_supplier.csv

I think either can be generated from the other. I think it's confusing to have 2, though, because from the names it sounds like they are not equivalent but that one is a more simple version of the other. And I think we would want people to work with the one with fewer columns. Otherwise, they have to manually track dependencies (like that utility is a function of value and automation). Which I think we want to avoid asking people to do. 

So I'm in favor of eliminating it.

j--- (changes requested):
Besides the one comment above, these all look good to me.

ahouseholder:
5ec0f40b6f2ea1cacd44026d874a12fe67fa9900 addresses the supplier tree file changes above.

j--- (approved):

sei-vsarvepalli:
There is a small issue with Javascript library for SSVC, that assumes the first column to be a numeric field that is ignored. You can see function parse_file().  

https://github.com/CERTCC/SSVC/blob/e024e5a5d5e498abd7e01c7a65519d9c15ae93b5/ssvc-calc/ssvc.js#L478 

thinking rn about best way forward.  The code might get messy if it needs to detect in CSV if is it version v1 or version v2.  I suggest leaving the column "row" 

Vijay

ahouseholder:
So the row numbers are basically a historical accident, I'm a little hesitant to maintain them just to 

I think options include:

1. Add v1/v2 discriminator in js, row numbers in v1, none in v2
2. Make js compatible with v2, break backwards compat with v1 trees (but fix by modifying v1 trees to drop rowid col)
3. Add row numbers to every v2 csv, same js handles both v1 and v2 trees
4. Have different versions of the js for v1 and v2 compatibility (freeze the v1 version to the csvs)

My preference would be for 2 rather than 1 or 3, assuming we need to have one js handle both v1 and v2 CSVs.
For the python lookup function, I'm planning on doing option 4. See #135.

sei-vsarvepalli (changes requested):
Yeh We need to updated SSVC JS script to change two things. Line 479 and Line 500 to be redone to detect v1/v2 to decide removal of ID column.

sei-vsarvepalli:
> Yeh We need to updated SSVC JS script to change two things. Line 479 and Line 500 to be redone to detect v1/v2 to decide removal of ID column.

Looks like my changes did not get added, may be because the files are outside the currently changed files. You can just go ahead and merge. I will create a new small issue to update the JS files.

PR #137, opened 2021-03-22
Use simple number search to detect v1 vs v2 CSV trees

sei-vsarvepalli:
Simple v1 or v2 detector script ensures that CSV files can optionally have Row-ID column, made of numbers.  The javascript checks to see if variable y[0][0] is a number, i.e. the first column element of the second row of CSV file (first row is assumed to be the header)

Vijay

### DIFF
--- a/ssvc-calc/ssvc.js
+++ b/ssvc-calc/ssvc.js
@@ -1,5 +1,5 @@
 /* SSVC code for graph building */
-const _version = 4.1
+const _version = 4.2
 var showFullTree = false
 var diagonal,tree,svg,duration,root
 var treeData = []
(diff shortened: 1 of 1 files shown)


sei-vsarvepalli:
I already pushed this branch to the production site at https://democert.org/ssvc/ it seems to detect both v1 and v2 CSV files through hat simple logic.  I know CISA uses the site regularly, but not sure if they import CSV - not possible to tell as we don't log JS calls.  Welcome to test it out.  

Vijay

ahouseholder (approved):
The demo deploy looks fine. I was able to import a v2 tree from the current main (post-merge of #134) so I think it should work.

PR #138, opened 2021-03-25
Add python lookup helper for v2

ahouseholder:
Here's a simple module that does the following:

1) Loads all the tree CSVs in /data/v2/csv into a dict of pandas data frames (one df per tree)
2) Provides a "lookup" method that, given a tree and query dict as arguments, will return a df containing all elements of the tree df matching the query. This allows for partial matches.
3) Provides an "outcome_dist" method that returns the distribution of outcomes across a df. Example usage would be when you have a partial query and want to know which outcome is most likely given what you know.

I don't think we have any formal requirements for what we need python helper code to do with the CSVs, so this is just intended as a demonstration usage. I can add some more simple things to this prototype code if we come up with anything here. However, I'd suggest that if we need significantly more features we should capture those in separate issues.


src/ssvc_v2.py
```diff
@@ -0,0 +1,113 @@
...
+
+        df = df.loc[df[k] == v]
+    return df
+
+def outcome_dist(df):
+    '''
+    Given a dataframe representing an SSVC tree fragment,
+    compute and return the distribution of outcomes
+    '''
+    return df['Outcome'].value_counts(normalize=True).to_dict()
```
sei-vsarvepalli, line 82: Perhaps round this panda frame before conversion to dict for ease of reading

``` suggestion
    return df['Outcome'].value_counts(normalize=True).round(2).to_dict()
```

sei-vsarvepalli (changes requested):
This is helpful for tree analysis itself. There are probably lot more use cases to analyze the SSVC tree and show some crisp representation of its properties and paths.

Vijay

src/ssvc_v2.py
ahouseholder: My position is that rounding is a display preference, and so that should be something that happens later in the caller of this method rather than here. 

See changes in 39677f5dcb9b3f0ee6c674b9ae032a62e1a167b2 for how I did this. And on reflection, the to_dict() is also a display choice, so I moved that out of the outcome_dist method as well.

sei-vsarvepalli (approved):
Looks good.  Once Jono reviews and I will be happy to merge and close this.

j--- (approved):
Thanks, this looks good to me (based on my code reading).

Issue #151, opened 2021-09-23
Items from 2021-09-23 meeting


ahouseholder:
Followup to #149 with a few more items from our conversation this morning.

Vector
- [ ] ISO 8601 dates in vector #150 
- [ ] Vector should include both "simple" and "complex" decision outputs, but without calling attention to which type they are

PDF
- [ ] PDF output should have "Recommendation" as the first item under Summary

JSON decision_points
- [ ] should allow for "simple" (no dependencies) or "complex" (which depend on answers to other decision points)
- idea: for "complex" decisions, the "options" structure could mirror the decision table structure e.g., a list of items such as `{ "situated safety impact": "Major", "mission impact":"MEF Failure", "output": "High"}`
- [ ] should be flat, not nested (i.e., `decision_points` is just a list of simple and complex things. The complex ones will need to refer to the simple things of which they are composed, but the simple ones don't need to say anything about how they show up in complex things. This should allow more composability without overly increasing complexity in the future.

ahouseholder:
If I've missed anything in the above, please either edit the description directly or leave a comment here with additions.

sei-vsarvepalli:
The current schema allows for a child identifier in the schema itself
https://github.com/CERTCC/SSVC/blob/e2d0f0a586a9c8bc9d06a1079cfcd937a52a5dce/data/schema/SSVC_Provision_v2.01.schema.json#L59-L66 

In the new setup, I will get rid of it or basically collapse "child" into "simple" as you guys suggested.  Comment if you "disagree" or want to change it.  

Currently tree structure also has the "children" which is the recursive decision points the way I explained. But I think the preference is to leave this flat.

https://github.com/CERTCC/SSVC/blob/e2d0f0a586a9c8bc9d06a1079cfcd937a52a5dce/data/schema/SSVC_Provision_v2.01.schema.json#L69-L73

If I change this, the children will only point to the child either by "shortcut key" decision_point.key or by full decision_point.label using anyOf in JSON schema setup. Just making sure the schema changes are  understood and digested.

sei-vsarvepalli:
[...]

Issue #150, opened 2021-09-23
Use ISO 8601  dates in vector representation instead of Unix Epoch Seconds


ahouseholder:
Equivalent to output of `date -u +"%Y-%m-%dT%H:%M:%SZ"`

sei-vsarvepalli:
Covered by PR #152

Issue #149, opened 2021-09-16
Requirements for new updates for the ssvc-calc tool aka Dryad


sei-vsarvepalli:
Here are the few new requirements from our sponsors on SSVC calc tool
1. Provide SSVC export in PDF format with a table and information
2. Export JSON to include full definitions of all the various options. 
3. Import CVE template for ssvc-calc to be available.
4. Display SSVC vector format in the screen right before Export
5. Provide permalink for people to share for the current decision for processing.

sei-vsarvepalli:
There seems to be a bug in the schema.json and json file examples where "options" has replaced "choices"  I can't remember if this was intentional or not?  may be @ahouseholder  or @j--- can recall? 

The code for ssvc.js has trouble due to this change in creating and assembling data fields on import and export. I will mark this also as a bug so either JSON+Schema or javascript should be modified or perhaps both?

Vijay

sei-vsarvepalli:
The move from "choices" to "options" was intentional and SSVC code is being updated to fix this bug. This came from #106 

There is a concern that SSVC Vector and the current export formats loose information that is collected as part of an SSVC decision that was made. For example as "Mission & Well-being" is a cumulative decision from two factors.  Current exported JSON format even with full tree only records the combinatorial decision from "M & WB" section.  We should find a way to preserve this data in both SSVC vector and in exported trees in JSON and PDF format.

Some possible solutions are 
1. For JSON - Provide an additional JSON blob that captures the decisions apart from the current "decisions_table" in JSON schema. Provide details of the sub-components that make a "complex" decision.
2. For PDF - include a Table for complex decision and include the current selections of an SSVC decision maker.

Current updates and features in the website https://democert.org/ssvc/

1. Two versions of PDF export are available, one with the full tree and documentation of the SSVC tree, and another that just has the current decision/recommendation reached.
2. There is a link in the PDF to the CVE if the vulnerability being scored is a CVE linking it to NVD.
3. SSVC score is shown in current vector presentation on the site before export and it is also included in the PDF itself.
4. Next to the SSVC score is a link icon that if you click will take you to the permalink URL showing the current decision as an animated path through the tree. Example - permalink example:  https://democert.org/ssvc/#SSVCv2/E:A/V:S/T:T/M:M/D:A/1632238698/&CVE-2014-0751&Coordinator
[...]

PR #152, opened 2021-10-01
All Schema v2.02 updates. Simplifying the code

sei-vsarvepalli:
This tries to address a number of Issues that we have been addressing 
Issue #149 => 
    Provide SSVC export in PDF format with a table and information
    Export JSON to include full definitions of all the various options.
    Import CVE template for ssvc-calc to be available.
    Display SSVC vector format in the screen right before Export
    Provide permalink for people to share for the current decision for processing.
Issue #150 
  ISO 8601 a specific version of the format that excludes "milliseconds" and uses the Z to represent UTC time
Issue #151 
Simplify the decision_points schema to be flat (only "complex" and "simple" no need for children identify themselves as "child" and to identify their parents.
Complex decisions can include other "complex" decisions. It will be recursive as the schema itself does not really care about. These logic should be put into the code that analyzes the tree JSON.


j--- (commented):
I think this looks OK.

PR #157, opened 2022-02-18
Removing hard-coded final keyword and final outcome

sei-vsarvepalli:
A minor update is needed so the the final step in the decision tree is not hard coded to be called "Decision" but can be other words that are actually read from the JSON file that has the decision tree adhering to SSVC JSON schema.

That is the only change in this commit, but essential for the CVE-5 ADP Pilot so it is possible to send ADP tree to CVE JSON schema folks to validate.

Vijay


ahouseholder (approved):
seems reasonable to me

Issue #160, opened 2022-03-15
Update Trees at democert


tschmidtb51:
Please update the trees for developers and appliers at https://democert.org/ssvc to reflect the current state.

sei-vsarvepalli:
Supplier tree has been added to the demo site.  The CSV files are also available for deployer, however the JSON format and the website content is pending at this time.   I will need a few more days to do a review of the various decision trees to consolidate some of the tree definitions and their relationships. 

Vijay

j---:
Static latex deployer tree also needs to be updated to s/Mission and Well-being Impact/Human Impact/g as per the document. That's a todo for me.

sei-vsarvepalli:
Hello @tschmidtb51 

The Deployer tree is also available at the demo site. Feedback welcome

https://democert.org/ssvc/

Thanks
Vijay

sei-vsarvepalli:
closed by #174

PR #161, opened 2022-03-15
recreated CSV files and added a folder for them with readme; updated generation scripts

j---:
contributes to #160 

I used the PDFs in doc/graphics to type in the priority labels. Hopefully there are no typos in the priority labels -- I don't see any. The decision point combinations are programmatically generated using the enumerate*.sh scripts, so the names for the decision point values are consistent since the script creates those.

### DIFF
--- a/data/csvs/README
+++ b/data/csvs/README
@@ -1,0 +1,8 @@
+These delimter separated lists are for convenience.
+If there is a conflict with an associated JSON schema, the JSON takes precedence.
+
+A skeleton for filling in a csv table can be generated using ../src/enumerate*.sh
+That is mostly useful if you want to customize your risk posture by changing the priority for some situations.
+
+These CSV files are used as input to ../src/SSVC_csv-to-latex.py as part of the pipeline for creating the pretty static images of the trees used in the PDF document.
+
--- a/data/csvs/deployer-options_v2.csv
+++ b/data/csvs/deployer-options_v2.csv
@@ -1,0 +1,109 @@
+row,Exploitation,Exposure,Utility,Well-being and Mission Impact,Priority
+1,none,small,laborious,low,defer
+2,none,small,laborious,medium,defer
+3,none,small,laborious,high,scheduled
+4,none,small,laborious,very high,scheduled
+5,none,small,efficient,low,defer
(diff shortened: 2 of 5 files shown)


sei-vsarvepalli:
There seems to be few folders in the shell scripts and python scripts.

`"../data/deployer-options_v2.csv"` in enumerate-deployer-options.sh
`"../data/csvs/deployer-options_v2.csv"` in enumerate-deployer-options.sh
`"../data/v{1,2}/csv/"` in the python scripts for v1 and v2 respectively

so we can standardize all to be `"../data/csvs/"` as you already have the `_v2` or `_v1` in the filename itself. 

 Vijay

sei-vsarvepalli:
A quick mockup of the script to be something like below - we could also generalize all of the filenames, version numbers to variables to avoid mistakes.. 

```
#!/bin/bash

# row numbers make change discussion a lot easier
i=1

# default output file in the SSVC Github file structure
out="../data/csvs/supplier-options_v2.csv"

# refuse to clobber existing file
if [ -e "$out" ] 
  then echo "File $out exists. Will not clobber. Exiting."
  exit 1
fi

#header row
decisions="Priority,defer,scheduled,defer,out-of-cycle,scheduled,out-of-cycle,scheduled,out-of-cycle,scheduled,out-of-cycle,scheduled,out-of-cycle,scheduled,out-of-cycle,scheduled,immediate,scheduled,immediate,out-of-cycle,immediate,out-of-cycle,immediate,out-of-cycle,immediate,out-of-cycle,immediate,out-of-cycle,immediate,out-of-cycle,immediate,out-of-cycle,immediate,immediate,immediate,immediate,immediate"
IFS=',' read -r -a decision <<< "$decisions"
echo "row,Exploitation,Utility,Technical Impact,Public-Safety Impact,${decision[0]}">$out
for Exploitation in none PoC active
  do for Utility in laborious efficient "super effective"
    do for Technical in partial total
      do for PublicSafety in minimal significant
          do echo $i,$Exploitation,$Utility,$Technical,$PublicSafety,${decision[i]}>>$out 
             i=$(($i+1))
      done
    done
  done
done
```

PR #170, opened 2022-03-29
propagate change in markdown to deployer image; prepare CSVs for sub-trees

j---:
The deployer tree was using "mission and well being impact" whereas the text in v2 changed the name for that concept to "human impact". This change was not propagated to the tree image. 

Separately, prepare CSVs that define the complex or derived decisions. Since these are a function of other decision points, make that clear in a way that can support automation.


sei-vsarvepalli:
It looks like also in `Mission Impact` the choices `None / Non-Essential Degraded` or `MEF Support Crippled `  doesn't seem to matter for Combined value of **Human Impact**

Value| Definition
-- | --
None / Non-Essential Degraded | Little to no impact up to degradation of non-essential functions; chronic degradation would eventually harm essential functions
MEF Support Crippled | Activities that directly support essential functions are crippled; essential functions continue for a time

https://github.com/CERTCC/SSVC/blob/main/doc/md_src_files/055_decision-points_2.md#mission-impact

Can we just collapse these two choices to one?   There may be other reasons to keep these separate - please comment.

sei-vsarvepalli (commented):
Some issues and concerns
1. The `Mission Impact` the choices `None / Non-Essential Degraded` or `MEF Support Crippled` doesn't seem to matter for Combined value of `Human Impact` Can we collapse these choices.
2. The file `data/csvs/mission-and-well-being_v1.csv` seems to be named with v1 - Is that a mistake?
3. Remove the file `data/csvs/public-safety-impact_v2.csv`  as it is just a one to one mapping that is not really a Cumulative score for decision point.

j---:
Thanks. 
1. The line in the table for Human Impact has them combined.

sei-vsarvepalli:
> Thanks.
> 
> 1. The lines in the table for Human Impact has them combined.
>    `Hazardous | None/Degraded/Crippled | High`
>    I expanded them in the CSV so the JSON could use a literal string match rather than have to start pulling apart lines in the CSV on slashes `/` . Is that wrong? We are taking those Mission Impact terms / levels from FEMA documentation in Federal Continuity Directive 1 and 2, so I would prefer to preserve the terms as they are. Any discussion on changing terms should go under material revision questions for v3, I think.
> 2. Yes that is on purpose, it uses SSVC v1 definitions in that the decision point is called "well-being impact" rather than the v2 term "safety impact"
> 3. If you don't need it, we can remove it.

On (3)-> I can clean up the not needed files.

I have uploaded the Deployer v2 tree to the demo site. Please take a look and see if it works as expected. The FEMA definitions of multiple `Mission Impact` statement will be maintained as-is.  It is possible these will need to be revisited as they do not have distinct definitions in our documentation and have been carried over into the web UI with the same "help" info mouseover data.

`None` => Little to no impact up to degradation of non-essential functions; chronic degradation would eventually harm essential functions

`Degraded` => Little to no impact up to degradation of non-essential functions; chronic degradation would eventually harm essential functions

Issue #169, opened 2022-03-29
ssvc-calc: Add coordinator and vendor trees


zmanion:
Low priority, add trees/selection options for
* Supplier 2.0
* Coordinator Triage 2.0
* Coordinator 2.0

sei-vsarvepalli:
This is for the demo site? Supplier 2.0 already exists in the demo and also in Github https://github.com/CERTCC/SSVC/blob/main/ssvc-calc/Supplier-v2.0.0.json 

The following trees are pending for the demo site 
- Triage
- Deployer

zmanion:
Sorry yes, the request is basically for the demo (and eventually "production") public calculator to support all of the trees: coordinator triage, coordinator publish, CISA coordinator, supplier, deployer.

sei-vsarvepalli:
The Deployer tree has been added the demo site. Pending some validation it will get added in the next Pull Request. 

https://democert.org/ssvc/Deployer-v2.0.0.json

sei-vsarvepalli:
More trees added to demo site as requested by @zmanion 

https://democert.org/ssvc/Coordinator-Triage-v2.0.0.json for Coordination centers like CERT/CC trying to decide whether to Triage a vulnerability report that has been submitted to them.

https://democert.org/ssvc/Coordinator-Publish-v2.0.0.json for Coordination centers like CERT/CC to pursue public notification or a public announcement of a vulnerability that is coordinated.

Note: These Coordination decisions may not be relevant to all types of CERT coordination centers.  Some CSIRT's may find this useful https://www.first.org/standards/frameworks/csirts/csirt_services_framework_v2.1

sei-vsarvepalli:
Closed by #174

Issue #167, opened 2022-03-28
Provide ssvc-calc code in no graphics mode for integration into third party websites. 


sei-vsarvepalli:
Provide the ability to integrate ssvc-calc without D3 tree graph libraries and any of the additional capabilities like export (PDF, JSON, PNG) for the calculator.  May have optional capabilities like Mouseover Events trigger for instructions that are available in the JSON schema file.

sei-vsarvepalli:
Closed by #174 the big PR for all these

PR #174, opened 2022-04-01
Tree updates and code update to fulfill request and recent issues

sei-vsarvepalli:
This PR should resolve a number of issues as related to enhancement, demo-site and updates to code to support non-graphic mode operation.

Issue #169,  #160  #167 should be addressed with this PR.

Thanks
Vijay


doc/md_src_files/060_decision-trees.md
```diff
@@ -23,31 +23,31 @@ How this decision information might be stored or communicated is the topic of su
...
+The example supplier tree [PDF](../graphics/ssvc_2_supplier.pdf) shows the proposed prioritization decision tree for the supplier. Both supplier and deployer trees use the above decision point definitions. Each tree is a compact way of expressing assertions or hypotheses about the relative priority of different situations. Each tree organizes how we propose a stakeholder should treat these situations. Rectangles are decision points, and triangles represent outcomes. The values for each decision point are different, as described above. Outcomes are priority decisions (defer, scheduled, out-of-cycle, immediate); outcome triangles are color coded:
 
   - Defer = gray with green outline
   - Scheduled = yellow
   - Out-of-Cycle = orange
   - Immediate = red with black outline
 
-![Suggested Supplier Tree](graphics/ssvc_2_supplier.pdf){ width=100% }
+![Suggested Supplier Tree](../graphics/ssvc_2_supplier.pdf){ width=100% }
```
j---, line 33: IIRC, for security reasons you can't include directory traversal in what will eventually get rendered as HTML links. So I don't think it will work for turning the markdown into a PDF. 
This isn't documented as a design requirement for the directory structure (apologies), but I believe everything needed to create a PDF should be in the `doc` directory, otherwise the pandoc scripts break. 
I can add this to `doc/style-guide-how-to` and/or the README? Unless you're able to run `doc/compile-citeproc.sh` with this reorganization and get a valid HTML file with no errors?

doc/md_src_files/060_decision-trees.md
```diff
@@ -23,31 +23,31 @@ How this decision information might be stored or communicated is the topic of su
...
+The example supplier tree [PDF](../graphics/ssvc_2_supplier.pdf) shows the proposed prioritization decision tree for the supplier. Both supplier and deployer trees use the above decision point definitions. Each tree is a compact way of expressing assertions or hypotheses about the relative priority of different situations. Each tree organizes how we propose a stakeholder should treat these situations. Rectangles are decision points, and triangles represent outcomes. The values for each decision point are different, as described above. Outcomes are priority decisions (defer, scheduled, out-of-cycle, immediate); outcome triangles are color coded:
```
j---, line 26: This was a full HTML reference on purpose to make the document portable. It's not an include for the image, but a reference to it. So it needs to stay a fully specified HTML link.

doc/md_src_files/060_decision-trees.md
```diff
@@ -73,13 +73,13 @@ Other coordinators should consider customizing the tree to their needs, as descr
 
 ### Publication Decision Tree
 
-Suggested decision values for this decision are available in [CSV](https://github.com/CERTCC/SSVC/blob/main/data/ssvc_2_coord-publish.csv) and [PDF](https://github.com/CERTCC/SSVC/blob/main/doc/graphics/ssvc_2_coord-publish.pdf) formats.
+Suggested decision values for this decision are available in [CSV](../../data/csvs/ssvc_2_coord-publish.csv) and [PDF](../graphics/ssvc_2_coord-publish.pdf) formats.
```
j---, line 76: ditto above about being a reference, not an include.

doc/md_src_files/060_decision-trees.md
```diff
@@ -23,31 +23,31 @@ How this decision information might be stored or communicated is the topic of su
...
   - Scheduled = yellow
   - Out-of-Cycle = orange
   - Immediate = red with black outline
 
-![Suggested Supplier Tree](graphics/ssvc_2_supplier.pdf){ width=100% }
+![Suggested Supplier Tree](../graphics/ssvc_2_supplier.pdf){ width=100% }
 
 <!-- This isn't portable
-<embed src="graphics/ssvc_2_supplier.pdf" alt="Suggested supplier tree" type="application/pdf"
+<embed src="../graphics/ssvc_2_supplier.pdf" alt="Suggested supplier tree" type="application/pdf"
```
j---, line 36: ditto above about breaking document compilation. 
I'm not going to put a note on the following changes, if this is the problem, but I think all following changes in this file have this problem unless otherwise noted.

doc/md_src_files/060_decision-trees.md
```diff
@@ -23,31 +23,31 @@ How this decision information might be stored or communicated is the topic of su
...
 style="width: 100%;"
 height = "600" />
 -->
 
 
 ## Deployer Tree
 
-The example deployer tree [PDF](https://github.com/CERTCC/SSVC/blob/main/doc/graphics/ssvc_2_deployer_SeEUMss.pdf) is depicted below.
+The example deployer tree [PDF](../graphics/ssvc_2_deployer_SeEUMss.pdf) is depicted below.
```
j---, line 44: Reference, not include, needs full HTML link.

ssvc-calc/CISA-Coordinator-v2.0.3.json
```diff
@@ -64,17 +64,17 @@
 		{
 		    "label": "Minimal",
 		    "key": "M",
-		    "description": "Type of harm is \"All\" (Physical, Environmental,Financial,Psychological). The effect is below the threshold for all aspects described in material."
+		    "description": "Type of harm is \"All\" (Physical, Environmental,Financial,Psychological) and the associated Safety Impact Value is \"None\". The effect is below the threshold for all aspects described in material."
```
j---, line 67: I think this should say:
For any type of harm (Physical, Environmental,Financial,Psychological) the Safety Impact value is \"None\" or \"Minor\".
This also means you can remove the trailing "The effect is below..." sentence because the description for material just says the value of safety impact is Major, and None/Minor fully captures what it means for the value to be below major.

ssvc-calc/CISA-Coordinator-v2.0.3.json
```diff
@@ -64,17 +64,17 @@
...
+		    "description": "Any one or more of the conditions (Physical, Environmental,Financial,Psychological) hold and the associated Safety Impact Value is \"Major\". \"Physical harm\" means \"Physical distress or injuries for users of the system OR introduces occupational safety hazards OR reduction and/or failure of cyber-physical system’s safety margins.\" \"Environment\" means \"Major externalities (property damage, environmental damage, etc.) imposed on other parties.\" \"Financial\" means \"Financial losses that likely lead to bankruptcy of multiple persons.\" \"Psychological\" means \"Widespread emotional or psychological harm, sufficient to be cause for counselling or therapy, to populations of people.\""
```
j---, line 72: This edit is good.
However it strikes me as a place for improvement, the descriptions of physical harm, etc., should probably be represented better. It seems weird to have them only on Material and Irreversible and not on Minimal. Can they be collected on just the description of the decision point, and not the options?

j--- (changes requested):
I believe the reorganization of files will cause some breakage. 
The content of the schema and json generally look good, but a couple small suggested changes.

ssvc-calc/CISA-Coordinator-v2.0.3.json
sei-vsarvepalli: Looks good. I will update the code to show Complex Decision points with required additional detailed information and hide it in the case of Options for the mouseover events.

PR #175, opened 2022-04-02
Corrected the "Suggested"-Hyperlinks

fruehaufm:
The hyperlinks are working now, but the PDFs aren't displayed.

### DIFF
--- a/doc/md_src_files/060_decision-trees.md
+++ b/doc/md_src_files/060_decision-trees.md
@@ -30,7 +30,7 @@ The example supplier tree [PDF](https://github.com/CERTCC/SSVC/blob/main/doc/gra
   - Out-of-Cycle = orange
   - Immediate = red with black outline
 
-![Suggested Supplier Tree](graphics/ssvc_2_supplier.pdf){ width=100% }
+![Suggested Supplier Tree](../graphics/ssvc_2_supplier.pdf){ width=100% }
 
 <!-- This isn't portable
 <embed src="graphics/ssvc_2_supplier.pdf" alt="Suggested supplier tree" type="application/pdf"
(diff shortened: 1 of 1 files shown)


j--- (changes requested):
I think I understand the confusion. from the perspective of the md file locations, the graphics dir is one level up. 

However, the relevant perspective is from where the `doc/compile-citeproc.sh` script is / is run from. That's `doc`. That's embedded in that script by the line `src="./md_src_files" `. However, I admit that no where is this obvious.

Can you help me understand what the problem was with the links prior to this PR? I don't see a linked issue that helps me understand what the old form of the links was breaking. It's possible this is the right thing to do, and the compile script needs to change, but I need some help understanding what we're aiming to fix, so we get it fixed everywhere.  

It's also possible that we should replace the `compile-citeproc.sh` script with a version that goes to PDF, rather than HTML. I think I got that working to embed the graphics files, so the relative links only need to work at compile time, locally, which would probably be more stable and easier to test. I'll get a version that goes to PDF rather than HTML committed.

fruehaufm:
I see, thanks for the clarification.

From the perspective of the md file, the links don't work. But they work fine for the perspective of `doc/compile-citeproc.sh`. Maybe this is just a misinterpretation of mine.

The problem for me was, that these links didn't work when displayed in a browser. It was linked as `Suggested Supplier Tree{ width=100% }` what didn't work from the perspective of the md file. That's why I proposed to change it to `Suggested Supplier Tree{ width=100% }`.

j---:
Mmm, yes, that problem makes sense. 
IIRC, originally I had the graphics folder inside md_src_files so that the compile script and the browser could reasonably run from the same directory location. There was separate feedback that no one could find the graphics when browsing the files with this set up, so they moved up a level in the hierarchy. However, as you point out, this breaks this consistency between the script and the browser. 

I would suggest the issue is to restore a consistent perspective for the compile script and the browser view? 
If that is the problem, I think this PR won't solve it, and we would have to make a new issue to decide what to do about it. 

Is that a correct description of the problem?

fruehaufm:
Okay, I get it.
This makes sense as well..

I think your description sums it up very well.
You're right. With my suggested PR the inconsistencies are only moved and affect the compile script instead of the browser view of the md files. So you're right. There should be a consistent perspective for the browser view and the compile script.
Sorry for only moving this issue from the browser to the script with my PR. I didn't know about this dependencies.

j---:
Can I close this PR without merging and we can address the problem via #181 ?

fruehaufm:
Sure, this problem can be closed and handled via #181 .
Thank you.

PR #180, opened 2022-04-06
Pdf update

j---:
This just provides the tools to create a new PDF from the markdown. Before creating a new PDF, want to check on the version numbering used in line 25, which is major.minor_date rather than proper semver, because then we don't have to make the script track git commits to the markdown in the doc folder. There is no special formatting to the PDF, it is a basic KOMA report template. 

I'd also like to make sure somone else can run the pandoc script. I think the requirements are pandoc and texlive. How should these be documented? 

If we can get it down to being as simple as running the script, we can ask for an updated PDF with every commit that changes the doc directory. 

Once this is settled, then we can have a different branch for adding an updated PDF.

### DIFF
--- a/doc/compile-html-citeproc.sh
+++ b/doc/compile-html-citeproc.sh
@@ -1,0 +1,24 @@
+#!/bin/sh
+src="./md_src_files"
+
+pandoc --self-contained \
+  --from=markdown_github+citations+table_captions+implicit_figures+link_attributes  \
+  --to=html \
+  --citeproc \
+  --bibliography="$src/sources_ssvc.bib" \
+  -M title="Prioritizing vulnerability response: A stakeholder-specific vulnerability categorization (DRAFT version 2.0)" \
+  -T "SSVC" \
+  -M author="Jonathan M. Spring; Eric Hatleback; Allen D. Householder; Art Manion; Madison Oliver; Vijay Sarvapalli; Deana Shick; Laurie Tyzenhaus" \
+  -M date="Compiled `date -u`" \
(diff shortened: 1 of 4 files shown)


ahouseholder:
Here's what I get when I try to run them:
```sh
$ ./compile-citeproc.sh
[WARNING] Deprecated: markdown_github. Use gfm instead.
$ ./compile-html-citeproc.sh 
[WARNING] Deprecated: markdown_github. Use gfm instead.
```

Pandoc info:
```
$ pandoc --version
pandoc 2.12
Compiled with pandoc-types 1.22, texmath 0.12.1.1, skylighting 0.10.4,
citeproc 0.3.0.8, ipynb 0.1.0.1
```

Other suggestions:
- Please add instructions on which scripts to run, in which order, to the `doc/README.md`. 
- I think the `doc/README.md` is also the place to note prerequisites (preferably with minimum versions included)
- Not sure if it helps your thinking on versioning, but https://calver.org/ talks about calendar versioning, and it does seem to have some uptake for folks where semantic versioning doesn't quite work as well.

sei-vsarvepalli (approved):
The updates work fine. We should look at long term @ahouseholder suggestions for GitHub actions and update to pandoc.  It looks like the `gfm`- the new GitHub markdown plugin has other incompatibility issues with `citations` from a quick search.  

So may need to rewrite the shell scripts. If the PDF generation is often, every pull-request or every quarter more than once, automating it definitely make sense.  It not, may be it is a bit overkill.

Vijay

sei-vsarvepalli:
Just merging this to close on it..

PR #187, opened 2022-05-20
Fixed redundant option for Mission Impact

fruehaufm:
According to the documentation on https://github.com/CERTCC/SSVC/blob/main/doc/md_src_files/055_decision-points_2.md#mission-impact "none" and "degraded" are actually just one option called "None/ Non-Essential Degraded".

### DIFF
--- a/ssvc-calc/Deployer-v2.0.0.json
+++ b/ssvc-calc/Deployer-v2.0.0.json
@@ -243,8 +243,8 @@
             "key": "M",
             "options": [
                 {
-                    "label": "none",
-                    "key": "N",
+                    "label": "degraded",
+                    "key": "D",
                     "description": "Little to no impact up to degradation of non-essential functions; chronic degradation would eventually harm essential functions. (aka Non-Essential Degraded)"
                 },		
                 {
@@ -297,7 +297,7 @@
 			      "child_label": "Mission Impact",
 			      "child_key": "M",
 			      "child_option_labels":[
-				  "none",
+				  "degraded",
 				  "crippled"
 			      ]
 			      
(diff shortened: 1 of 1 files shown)


sei-vsarvepalli:
Hi Michael @fruehaufm 

We left this as duplicate on purpose, there may be cases where this can be made granular differentiated. Eventually these may collapse like you suggested. I understand it may be confusing at this time. Perhaps we can update the documents to indicate this?  Let me know if that will help.

Thanks
Vijay

fruehaufm:
Hi @sei-vsarvepalli 
Thanks for your reply!
I see. If this duplicate is left on purpose, you just need to update the documentation and the description of the decision points. Right now I think it's a bit unclearly described.

Thanks
Michael

j---:
I think one move I'd like to make overall is to move away from saying "none" because that can be interpreted as categorically deny there is *any* impact whatsoever. This is probably never strictly true. Or at least, if the vul does not exist in the organization's infrastructure at all, that should just be a pre-filter on the vul should not be considered. If the intent is to capture the vul may be present and you don't know for sure it is not, then "none" is not appropriate, anyway. 
So I think I'm OK with saying "degraded or less" in the documentation for degraded. If that's what we agree to, then we should remove "none" from the list of options, not "degraded". 

If that's what we agree, there should be one Pull removing "none" from the schema and also updating the documentation to make that clear.

j---:
> That sounds plausible to me.

OK, great. Do you want to edit your pull to reflect that?

fruehaufm:
I just edited it and removed "none" instead of "degraded".

j--- (approved):
I think this covers everything.

sei-vsarvepalli (approved):
All checks out correct.  There seems to be no issue with this PR, approved to merge.

PR #205, opened 2022-11-01
fixed a typo

2shiori17:
The version number in the file name and the URL seem to differ.


j---:
@sei-vsarvepalli , are there any other places where this version # in the URL needs to be updated?

sei-vsarvepalli (approved):
All changes looks right thanks @j---  for expanding the mismatch in version numbers.

PR #231, opened 2023-04-28
add github workflow to generate html and pdf artifacts 

### LINKED ISSUE #184: Automate PDF generation with github actions
Once PR #180 is resolved, we should create a github action that can be triggered in a few different ways, including:

- [ ] on-demand on a particular branch
- [ ] automatically as part of the pulll request process

This will require (at least)

- [ ] installing requirements
- [ ] running the script
- [ ] commit and add the generated PDF

ahouseholder:
resolves #184 

This PR includes the following changes:

- a github workflow that will produce an HTML and PDF artifact on each push to `main` or `develop` branches that affects any file in the `doc` directory or the workflow file itself.
- adds a `Makefile` that replaces previous shell scripts for manually generating HTML and PDF locally (converting these scripts to a `Makefile` helped with figuring out the structure of the github workflow) 
    - sample output from the `Makefile` can be found in `ssvc_v2-0.html` and `ssvc_v2-0.pdf` 
- Updates some of the "how-to" files in the `doc` directory to markdown format

Note that the github workflow uploads its artifacts to the workflow summary page rather than modifying the repository directly.

This PR is a squashed version of my current `develop` branch, so if you want to see a sample of the output of the workflow you can check out https://github.com/ahouseholder/SSVC/actions/runs/4834190249


sei-vsarvepalli (approved):
Wow - Good work @ahouseholder. I will follow-up with Dryad (ssvc-calc folder) updates to do the same and publish these here instead of democert.org using GitHub Pages.

PR #247, opened 2023-06-16
Remove version strings from file names

### LINKED ISSUE #246: Stop including version numbers in file names
Over in #238 I came across a concern about the use of version numbers in file names.

- `.csv` files in [data/csvs](../blob/main/data/csvs) have strings like `v1`, `v2`, `v2_1` 
- `.json` files in [data/schema](../blob/main/data/schema) have strings like `v2.03`
- `.json` files in [ssvc-calc](../blob/main/ssvc-calc) have strings like `v2.0.3`, `v2.1.0` 

Regardless of our versioning strategy, I propose that we eliminate version numbers from file names.

Rationale:
- Either we strictly enforce that _every_ edit to a file _must_ change the file name _or_ a file will be edited more often than its name will change.
- This will mean that over time the repository will contain multiple versions of a  file named `foo-vX.Y.Z.json`, which has side effects, including:
  - the version string in the name is actually misleading because the file changed but the identity defined by the version string did not
  - sometimes the content of that file will serve as the basis for a new file `foo-vA.B.C.json` and `foo-vX.Y.Z.json` will be removed.
  - when changing the file name, a developer may be concerned about preserving the `X.Y.Z` version and have the urge to  create an `archive` folder or some such
  - while the preservation/conservation concern is warranted, by creating archive files we'd be double-version-controlling. One using git, the other using human copy-to-folder process.
- I'd rather trust git to be consistent in this process.

The alternative I'm proposing is that we remove version numbers from file names and let the git commit history be the canonical source for older versions of those files. So if you want to see the `foo-v1.json`, go to the v1 tag in the history and get it. My preference is to let git be git, and avoid overthinking version strings in file names.

ahouseholder:
resolves #246


cgyarbrough (approved):
I concur and approve these changes. I presume that the SSVC Provision Table would change with each customer, correct? I know there is a CISA-specific component for this run.

ahouseholder:
> I presume that the SSVC Provision Table would change with each customer, correct? I know there is a CISA-specific component for this run.

We include the CISA tree because we had it from helping to develop https://www.cisa.gov/ssvc-calculator (which is based on the ssvc-calc included in this repo.

However, it strikes me that the default tree that loads for _our_ calculator should probably _not_ be the CISA-tuned one, but instead one of the ones described in the document. I'll create a separate issue to capture that.

laurie-tyz:
@ahouseholder & @cgyarbrough we need to establish our own tree.

PR #250, opened 2023-06-21
Update 055_decision-points_2.md

j---:
Improve text description for Mission Impact.

There's not an associated issue with this, reflects some private feedback I received about better aligning with relevant documentation.


cgyarbrough (approved):
This is fine as submitted, but have we considered the DOD's MEF definition and how it aligns with CISAs concept? I understand that the current context is CISA-based so this is appropriate. Just a question--not a show stopper.

j---:
> This is fine as submitted, but have we considered the DOD's MEF definition and how it aligns with CISAs concept? I understand that the current context is CISA-based so this is appropriate. Just a question--not a show stopper.

The cited MEF documents are FEMA actually, so it's definitely a US federal civilian focus, but it's broader than just CISA. 

But anyway, it's a good point in general to add additional contexts, and the Defense use case is a good one. Do you want to suggest a sentence to add in the paragraph that notes the NCF and business continuity use cases as being distinct? Just add another paragraph about the relevant DoD documentation for mission continuity planning? 
Or is it sufficiently different that a defense sector couldn't use the information in a similar way?

ahouseholder:
@cgyarbrough check my Google digging, but I found:

> **MEFs**. Select functions directly related to accomplishing the Department’s mission. Failure to
perform or sustain these functions, which directly support PMEF, would significantly affect the
Department of Defense’s ability to provide vital services or exercise authority, direction, and
control.

...which refers to

> **PMEFs**. The DoD MEFs, validated by the National Continuity Coordinator, which must be
performed in order to support the performance of the national essential functions before, during,
and in the aftermath of an emergency. PMEFs need to be performed continuously or resumed
within 12 hours after an event and maintained for up to 30 days or until normal operations can be
resumed.

...and

> **national essential functions**. The select functions that are necessary to lead and sustain the
United States during a catastrophic emergency and that, therefore, must be supported through
COOP, COG, and ECG capabilities.

In [DOD DIRECTIVE 3020.26 DOD CONTINUITY POLICY](https://www.esd.whs.mil/Portals/54/Documents/DD/issuances/dodd/302026p.pdf)

Is that what you're referring to, or is there another definition elsewhere?

ahouseholder (approved):
This seems fine to me.

ahouseholder:
In the interest of merging this PR, I'm going to spawn a separate issue to address @cgyarbrough's comment above about DoD's MEF definition.

PR #256, opened 2023-06-28
add subsubsection header for tree versioning

ahouseholder:
The paragraph about tree versioning was buried inside the _Parsimony_ section, just calling it out with its own heading.


j--- (approved):
Good idea.

PR #258, opened 2023-06-30
Change default tree to Deployer.json

### LINKED ISSUE #257: The default ssvc-calc tree should be one described in the document
The ssvc-calc tool in this repo includes the CISA tree because we had it from helping to develop https://www.cisa.gov/ssvc-calculator. It also serves as a good example of tree customization for a specific stakeholder's needs.

However, it seems odd to me that _our_ calculator loads a custom tree by default instead of one of the ones described in the SSVC document itself.

Instead, I propose that the default tree that ssvc-calc loads should be one of the trees described in the SSVC document, namely one of:

- Supplier
- Deployer
- Coordinator-Triage
- Coordinator-Publish

Ranking these in order of potential audience size (and therefore addressing the most likely audience with the default), I believe the Deployer tree to be the most appropriate default tree to load.

_Originally raised in https://github.com/CERTCC/SSVC/issues/247#issuecomment-1614780011_

ahouseholder:
fixes #257 

Change the default tree loaded by `ssvc-calc` to the one described in `Deployer.json` instead of `CISA-Coordinator.json`


cgyarbrough (approved):
I concur with the changes. Deployer is a much more generic term.

PR #265, opened 2023-07-05
mention publicly known default passwords as example of Exploitation:PoC

### LINKED ISSUE #199: update *State of Exploitation* PoC def
Add as an example that a publicly-known but hardcoded or default password would count as a public PoC for the purposes of exploiting the vul (assuming the hardcoded password is itself the vul, which is consistent with CVE  usage).

ahouseholder:
fixes #199


cgyarbrough (approved):
I concur with this change. I like the more granular definition of PoC. I think this is clear.

j--- (commented):
I'm OBE, but looks good.

PR #270, opened 2023-07-07
Update analyze_csv.py to reflect csv column name changes

ahouseholder:
The column formerly known as "outcome" is now "priority" in the `.csv` files, so this change adjusts the default column it's looking for to match.


sei-vsarvepalli (approved):
checked!

Issue #249, opened 2023-06-21
Review new deployer tree's decision recommendations


j---:
> 41,PoC,open,no,low,defer

Consider making this scheduled.

> 54,active,small,yes,medium,scheduled

Consider making this out of cycle

> 61,active,controlled,yes,low,scheduled

Consider making this out of cycle

> 66,active,open,no,medium,scheduled

Consider making this out of cycle.

PR #276, opened 2023-07-11
Adjust Supplier Tree decisions

### LINKED ISSUE #275: Adjust supplier tree decisions 3 and 11
Based on discussion #259:

Change
```
3,none,laborious,total,minimal,defer
```
to
```
3,none,laborious,total,minimal,scheduled
```
and

```
11,none,super effective,total,minimal,scheduled
```
to
```
11,none,super effective,total,minimal,out-of-cycle
```

ahouseholder:
Fixes #275
Resolves #259 

Also includes a separate fix for `Deployer.json` that we missed in #262 related to #249


j--- (approved):
These are good. 
Just tracking that we need to rerun https://github.com/CERTCC/SSVC/blob/main/src/SSVC_csv-to-latex.py to update the graphics after this pull is accepted.

ahouseholder:
> Just tracking that we need to rerun https://github.com/CERTCC/SSVC/blob/main/src/SSVC_csv-to-latex.py to update the graphics after this pull is accepted.

I'm going to give this a look before I merge this PR in case I can just roll it in at the same time.

PR #281, opened 2023-07-14
Mention DoD 3020.26 MEF definition in Mission Impact

cgyarbrough:
These changes/additions express the DoD's concept of Mission Essential Functions, which is a slightly different use case than FEMA's which is what we original presumed.

### DIFF
--- a/doc/md_src_files/05_07_mission_impact.md
+++ b/doc/md_src_files/05_07_mission_impact.md
@@ -1,26 +1,31 @@
 ## Mission Impact
 > Impact on Mission Essential Functions of the Organization
 
-A **mission essential function (MEF)** is a function “directly related to accomplishing the organization’s mission as set forth in its statutory or executive charter” [@FCD2_2017, page A-1]. Identification and prioritization of mission essential functions enables effective continuity planning or crisis planning. Mission Essential Functions are in effect critical activities within an organization that are used to identify key assets, supporting tasks, and resources that an organization requires to remain operational in a crises situation, and so must be included in its planning process. During an event, key resources may be limited and personnel may be unavailable, so organizations must consider these factors and validate assumptions when identifying, validating, and prioritizing MEFs.
+A **mission essential function (MEF)** is a function “directly related to accomplishing the organization’s mission as set forth in its statutory or executive charter” [@FCD2_2017, page A-1]. 
+Identification and prioritization of mission essential functions enables effective continuity planning or crisis planning. 
+Mission Essential Functions are in effect critical activities within an organization that are used to identify key assets, supporting tasks, and resources that an organization requires to remain operational in a crises situation, and so must be included in its planning process.
+During an event, key resources may be limited and personnel may be unavailable, so organizations must consider these factors and validate assumptions when identifying, validating, and prioritizing MEFs.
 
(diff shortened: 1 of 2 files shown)


doc/md_src_files/05_07_mission_impact.md
```diff
@@ -7,6 +7,21 @@ When reviewing the list of organizational functions, an organization must first
...
+Department of Defense’s ability to provide vital services or exercise authority, direction, and
+control.
+
+To evaluate performance of DoD MEFs, validated by the National Continuity Coordinator, which must be
+performed in order to support the performance of the national essential functions before, during,
+and in the aftermath of an emergency. PMEFs need to be performed continuously or resumed
+within 12 hours after an event and maintained for up to 30 days or until normal operations can be
+resumed.
+
+The DoD's concept of national essential functions is more aligned with FEMA's notion. The select functions that are necessary to lead and sustain the
```
ahouseholder, line 21: So I started off working on adding a reference but in the process I went and looked at the FCD-2 doc to see why DoD and FEMA's perspectives were different -- especially since the concept of a National Essential Function seems like it'd be coming from outside DoD, and what I found makes me question whether we need this change at all. 

If you look at [FCD-2](https://www.fema.gov/sites/default/files/2020-07/Federal_Continuity_Directive-2_June132017.pdf) which is what we're referencing for the MEF definition, its framing seems to match the NEF -> PMEF -> MEF hierarchy that DoD outlines in 3020.26.

From FCD-2's _Purpose_:
>FCD-2 outlines requirements and
provides checklists and resources to assist D/As in identifying and assessing their essential
functions through a risk-based process and in identifying candidate PMEFs that support the
National Essential Functions (NEFs). This FCD provides guidance for conducting a BPA and
BIA to identify essential function relationships, dependencies, time sensitivities, threats,
vulnerabilities, consequences, and mitigation strategies related to the performance of the MEFs
and PMEFs. This FCD also provides direction on the formalized process for submitting D/As’
candidate PMEFs in support of the NEFs.

Which means that the assertion here that DoD's NEF matches FEMA's MEF is inaccurate. I think we need to reframe this either into a smaller change or just leave the doc as-is.

We seemed to be okay previously with just mentioning MEFs without situating them as the bottom of a hierarchy leading to NEFs, which I think makes sense because many organizations outside the USG could use the MEF concept even if they're not specifically required to. 

So I'm wondering if we actually just need to note here that DoD's MEF and FEMA's MEF are drawn from the same definitions, and indicate that they are part of a hierarchy of essential functions without going into much detail to avoid overfitting SSVC's concept (which is _borrowing_ from FEMA anyway) onto how USG defines things.

ahouseholder (changes requested):
See comment on individual change: As of this morning's reading, I think the assertion that DoD MEF != FEMA MEF is incorrect, and that they are actually exactly aligned. We can note that in a change to the doc, of course. But we shouldn't be asserting that they're different when it sure looks like they're the same. (I only just found this this morning when looking for NEF and PMEF terms in FCD-2, sorry I hadn't noticed that sooner.)

doc/md_src_files/05_07_mission_impact.md
ahouseholder: Resolved by commit 791dcabd716c2e681215493b26cba79f3863887b

PR #283, opened 2023-07-17
update pdf and html drafts

ahouseholder:
To reflect v2.1


ahouseholder:
Overriding review requirement because these are literally just docs generated from the existing repository.

PR #298, opened 2023-09-01
update authors, ack previous authors

ahouseholder:
Resolves discussion #280 
Fixes #297

Issue #274, opened 2023-07-11
Update pdf-html action to better integrate output artifacts


ahouseholder:
The github action workflow that generates a `.pdf` and `.html` file leaves its output in the individual run logs for the action itself. It would be better if these artifacts were made more accessible so we could have a consistent pointer in the documentation to the latest version of the rendered docs.

E.g.,
```
The latest version is [here](link_to_the_latest_pdf).
```

Based on [this comment](https://github.com/CERTCC/SSVC/issues/272#issuecomment-1630848545) this issue extends #272.

ahouseholder:
See for ideas https://joht.github.io/johtizen/build/2022/01/20/github-actions-push-into-repository.html#all-in-one-solutions

ahouseholder:
Given the change to the web site as the primary documentation target instead of PDF-building, I think this issue is now moot and can be closed as wontfix. I'm leaving it open for a bit longer to allow for counterarguments, but let's set a timer that it can be closed after 2023-10-13 if no such reasons are forthcoming.

PR #300, opened 2023-09-01
update draft docs

ahouseholder:
update based on latest doc build


ahouseholder:
this will be less hassle once we do #274

ahouseholder:
Manually overriding policy since this is just a direct follow-on effect of PR #298

j---:
Thanks!

PR #327, opened 2023-10-02
add drop column importance

### LINKED ISSUE #309: Add drop column importance to evaluation code?
A talk by Chris Madden @Crashedmind about theparanoids/PrioritizedRiskRemediation mentioned that Drop Column Importance might be preferable to (or at least augment) permutation importance, which we use in `src/analyze_csv.py`.

We should look into that.

### LINKED ISSUE #334: Add python test runner workflow
For pull requests against `main`, automate a test runner

ahouseholder:
This PR changes the default behavior of `analyze_csv.py` to use drop column feature importance instead of permutation importance. The permutation importance calculation is retained as the `--permutation` option on the command line.
Also refactored the module for unit testing and added a basic github workflow to run python unit tests.

- resolve #309 
- resolve #334


sei-vsarvepalli (approved):
looks powerful...

Issue #321, opened 2023-09-27
Update SSVC JSON schema to reflect versioned decision points and groups


ahouseholder:
Based on:

- #289 
- #303 

We need to revise the JSON schema to reflect the changes in the versioning discussion.

ahouseholder:
- blocks PR #328

PR #328, opened 2023-10-02
Implement python dataclass_json objects to represent decision points

### LINKED ISSUE #319: Add Critical Software decision point
I'm suggesting we add one (or more) decision points to reflect software criticality.

While I don't currently have a specific decision in mind for this to be used in, it seems likely to arise in stakeholder processes.

I am currently aware of two such Critical Software definition efforts:

1. NIST [definition](https://www.nist.gov/itl/executive-order-improving-nations-cybersecurity/critical-software-definition)
2. OSSF [working group](https://github.com/ossf/wg-securing-critical-projects)

Two approaches I can see to do this:
1. Treat the different definitions as a data mapping problem, and create a single, one-bit `Critical Software=(yes,no)` decision point. This implies documentation guidance that explains how to map NIST, OSSF, or any other such list onto the decision point. 
2. Treat each definition as its own distinct decision point, e.g., `NIST Critical Software=(yes,no)`, `OSSF Critical Software=(yes,no)` etc.  This eliminates the need for the data mapping portion of approach 1, but it leads to having more decision points addressing a fairly narrowly defined concept.

My opinion is that approach 1 is preferable, as approach 2 might be overfitting the solution to the problem. It's certainly possible to move from approach 1 to approach 2 in the future should we have a specific need to do so. But I don't think going from approach 2 to approach 1 is as easy because our decision point definitions are likely to become entrenched in a way that the data mapping approach seems more adaptable to.

This may or may not be the same as #210 which refers to High Value Assets (HVA), but we should consider both in tandem to figure out how many decision points we actually need. The part I don't understand is how do CS and HVA designations intersect in actual operational usage.

### LINKED ISSUE #325: Create python package to facilitate json round-tripping
Blocked by
- [x] #321 

Fill in more detail as needed, but at least:

- use `dataclasses_json` + `dataclasses` for the objects
- be able to generate json examples from python code to incorporate into reference documentation
- python installer for package using `pyproject.toml` https://pip.pypa.io/en/stable/reference/build-system/pyproject-toml/
- include code documentation in the reference docs as well

### LINKED ISSUE #330: Model CVSS v1, v2, v3 vector elements as decision points
Because of their ubiquitous availability in vulnerability data sets, we should provide set of decision points specifically to represent CVSS vector elements (aka _metrics_).

This would also provide us with an opportunity to further exercise our emerging decision point versioning rules, because CVSS has had more revisions over time than SSVC has.

The bulk of the work is in modeling CVSS v3 and v4, but the marginal effort to modeling CVSS v1 and v2 as well is small enough that we should just do all of them.

Edit: In order to line this up with PR #343, this issue is just going to cover v1, v2, and v3. v4 will be covered in a separate issue.

ahouseholder:
Starting this as a draft so others can look/comment

- [ ] Blocked by #321

- resolves #325 
- resolves #319 
- resolves #330


ahouseholder:
Resolves #325 
Resolves #328

ahouseholder:
Resolves #330

ahouseholder:
in https://github.com/CERTCC/SSVC/discussions/289#discussioncomment-7145569 @sei-vsarvepalli comments:

> 1. Looks like python3.10 is required for this version of dataclasses - the default python on Mac is 3.9 looks like. Should we mention this somewhere? Python3.9 throws an error for dataclass(kw_only=True) .
> 2. The https://github.com/ahouseholder/SSVC/blob/a254e9c926385049162c1f9284072d9dfab5c451/docs/reference/examples/human_impact.json#L9 calls the "values" output as "name" shouldn't this be labeled "value"? This we can disambiguate the "name" that is used for the Decision Point itself.
> 3. The Decision Point Group should not have any of the following fields https://github.com/ahouseholder/SSVC/blob/a254e9c926385049162c1f9284072d9dfab5c451/src/ssvc/dp_groups/coordinator_publication.py#L14C1-L17C21 - it is only a collection of decision points. These seem like they belong in the Decision Tree and not in a Decision Point Group as I understand from our discussions.

docs/reference/examples/human_impact.json
```diff
@@ -0,0 +1,29 @@
+{
+  "namespace": "ssvc",
+  "version": "1.0.0",
+  "name": "Human Impact",
+  "description": "Human Impact",
+  "key": "HI",
+  "values": [
+    {
+      "name": "Low",
```
ahouseholder, line 9: Comment from @sei-vsarvepalli:

> calls the "values" output as "name" shouldn't this be labeled "value"? This we can disambiguate the "name" that is used for the Decision Point itself.

src/ssvc/dp_groups/coordinator_publication.py
```diff
@@ -0,0 +1,40 @@
...
+created_at: 9/21/23 11:40 AM
+"""
+from ssvc.decision_points.exploitation import EXPLOITATION_1
+from ssvc.decision_points.public_value_added import PUBLIC_VALUE_ADDED_1
+from ssvc.decision_points.supplier_involvement import SUPPLIER_INVOLVEMENT_1
+from ssvc.dp_groups.base import SsvcDecisionPointGroup
+
+
+COORDINATOR_PUBLICATION_1 = SsvcDecisionPointGroup(
+    name="Coordinator Publication",
```
ahouseholder, line 14: Comment from @sei-vsarvepalli:

>The Decision Point Group should not have any of the following fields - it is only a collection of decision points. These seem like they belong in the Decision Tree and not in a Decision Point Group as I understand from our discussions.

ahouseholder: Per conversation yesterday, I refactored another mixin class `_Keyed` that is used by Decision Points and Decision Point Values but _not_ Decision Point Groups in https://github.com/CERTCC/SSVC/pull/328/commits/57609f1a8b7f7c597576b2fb9cd48685b22fa6a9
then removed the "key" assignments from the DP group objects in
https://github.com/CERTCC/SSVC/pull/328/commits/63a3ca11ccf4e71cc10566d238620eaf89d64793

`DecisionPointGroup` objects only have `name`, `description`, and `version` now. I think that tracks what we discussed, but let me know if I've missed something.

docs/reference/examples/human_impact.json
ahouseholder: My approach was to create a python base class (`ssvc._mixins._Base`) that just has `name` and `description` attributes. Both `SsvcDecisionPointValue` and `SsvcDecisionPoint` objects inherit from this `_Base`. I don't see a problem with classes consistently having a `name` attribute. It seems less desirable to me to have one object's name be in the `name` attribute and another object's `name` be in a `value` attribute. The class name already indicate that it's a `SsvcDecisionPointValue` so I think it would be unabmbiguous that an `SsvcDecisionPointValue.name` is the name of a decision point value whereas `SsvcDecisionPoint.name` is the name of a decision point.

sei-vsarvepalli:
Waiting on @sei-vsarvepalli for JSON format of JSON schema.

src/ssvc/_mixins.py
```diff
@@ -0,0 +1,58 @@
...
+    """
+    Mixin class for namespaced SSVC objects.
+    """
+
+    namespace: str = "ssvc"
+
+
+@dataclass_json
+@dataclass(kw_only=True)
+class _Keyed:
```
ahouseholder, line 34: This mixin class was added so that we can remove the key from the base class that DecisionPointGroup inherits from.

src/ssvc/decision_points/base.py
```diff
@@ -0,0 +1,82 @@
...
+from ssvc._mixins import _Base, _Keyed, _Namespaced, _Versioned
+
+
+## notes
+# based on https://gist.githubusercontent.com/sei-vsarvepalli/de2cd7dae33e1e1dc906d50846becb45/raw/2a85d08a4028f6dd3cc162d4ace3509e0458426f/Exploitation.json
+
+
+@dataclass_json
+@dataclass(kw_only=True)
+class SsvcDecisionPointValue(_Base, _Keyed):
```
ahouseholder, line 22: SsvcDecisionPointValues now inherit from _Keyed so they have a key.

src/ssvc/decision_points/base.py
```diff
@@ -0,0 +1,82 @@
...
+    """
+    Models a single value option for a decision point.
+    """
+
+    pass
+
+
+@dataclass_json
+@dataclass(kw_only=True)
+class SsvcDecisionPoint(_Base, _Keyed, _Versioned, _Namespaced):
```
ahouseholder, line 32: SsvcDecisionPoints now inherit from _Keyed so they have a key.

src/ssvc/dp_groups/coordinator_publication.py
ahouseholder: See comments beginning with https://github.com/CERTCC/SSVC/pull/328#discussion_r1352876222 below for specific changes to which classes have keys.

sei-vsarvepalli (approved):
Reviewed primarily the code for src/ssvc to use Python3.10 dataclasses and creation of JSON schema.  The JavaScript libraries and demo calculator is not updated as yet.

src/ssvc/decision_points/base.py
sei-vsarvepalli: For JSON Schema creation and automation. I am using the extraneous `field` method with a default metadata and an extension `schemaprops` all defined in base.py  as part of the setup. The _mixin.py also has the similar setup to help automate the creation of JSON schema doc

https://github.com/ahouseholder/SSVC/blob/89daecb74a6289c890dfa8799837b3304921f370/src/ssvc/_mixins.py#L19

https://github.com/ahouseholder/SSVC/blob/89daecb74a6289c890dfa8799837b3304921f370/src/ssvc/decision_points/base.py#L43-L54 

As dataclasses_jsonschema is a bit bloated for this application, our own method is created in doc_helpers/tools.py. I cannot see if we are using GitHub actions to create and automate all the files, so for now, I have just hand created then. The workflow can be updated so the docs/reference/schema/ is entirely automated and is not part of any manual creation. As we are unlikely to change this schema too often it is a bit more than we need to do at this time.

https://github.com/ahouseholder/SSVC/blob/89daecb74a6289c890dfa8799837b3304921f370/src/ssvc/doc_helpers/tools.py#L76

docs/reference/examples/automatable.json
j---, line 1: I don't know that the kill chain is such a ubiquitous concept that we should try to get away with the short version of the description not including a reference. What do we think? I know it adds a lot of length. And adding a reference to the bibliography entry for hutchins2011intelligence here would create an undesirable dependency. But can we add a link to https://www.lockheedmartin.com/content/dam/lockheed-martin/rms/documents/cyber/LM-White-Paper-Intel-Driven-Defense.pdf  ?

docs/reference/examples/exploitation.json
```diff
@@ -0,0 +1,24 @@
...
+  "description": "The present state of exploitation of the vulnerability.",
+  "key": "E",
+  "values": [
+    {
+      "name": "None",
+      "description": "There is no evidence of active exploitation and no public proof of concept (PoC) of how to exploit the vulnerability.",
+      "key": "N"
+    },
+    {
+      "name": "PoC",
```
j---, line 14: Maybe a separate issue -- Should we re-name this to "Public PoC" to emphasize it's not just a PoC used to validate that the vulnerability is a real vulnerability?

docs/reference/examples/exploitation.json
```diff
@@ -0,0 +1,24 @@
...
+      "description": "One of the following cases is true: (1) private evidence of exploitation is attested but not shared; (2) widespread hearsay attests to exploitation; (3) typical public PoC in places such as Metasploit or ExploitDB; or (4) the vulnerability has a well-known method of exploitation.",
```
j---, line 15: Do we actually think (1) and (2) are operationalizable? 
CISA's SSVC version of *State of Exploitation* removes them https://www.cisa.gov/sites/default/files/publications/cisa-ssvc-guide%20508c.pdf
Maybe we should move for consistency between instances of the decision point?

docs/reference/examples/human_impact.json
j---: FWIW, it would make sense to me if the elements of "values" were "value" and not "name" but as long as the name space is unambiguous, it's OK.

docs/reference/examples/mission_impact.json
```diff
@@ -0,0 +1,29 @@
+{
+  "namespace": "ssvc",
+  "version": "2.0.0",
+  "name": "Mission Impact",
+  "description": "Impact on Mission Essential Functions of the Organization",
```
j---, line 5: Similar question to my comment on *Automatable*, can we embed a link to the presidential policy directive here? https://www.cisa.gov/sites/default/files/2023-01/ppd-21-critical-infrastructure-and-resilience-508_0.pdf

ahouseholder: I don't want to have to bump a version in an SSVC object when a URL changes on a site we don't control. The description that goes into the json should be fairly minimalist. The documentation surrounding it can have whatever details we want to add.

docs/reference/examples/exploitation.json
ahouseholder: Separate issue.

docs/reference/examples/exploitation.json
ahouseholder: Separate issue.

docs/reference/examples/automatable.json
ahouseholder: SSVC object versions should not have to change to update links to external websites. The description field is not the documentation.

ahouseholder:
I closed this PR. I was not expecting commits to a branch in my repo and I had other unpushed work that was dependent on the branch as it was.

The changes made by @sei-vsarvepalli and @j--- are available in another branch now https://github.com/ahouseholder/SSVC/tree/feature/update_json_schema 

It's too disruptive to my work in progress to have commits directly to the branch I'm actively working on. If you want to clone the update_json_schema branch and do a PR against my pyjson branch, let's do it that way?

ahouseholder:
Reopening to keep the PR active, but (a) please don't push to this branch directly while it's in draft mode, and (b) there are a number of comments in the thread that should be separate issues entirely. This PR is about getting the python stuff in place. Content changes are out of scope for this PR.

docs/reference/examples/exploitation.json
ahouseholder: #353

ahouseholder:
Opened a few issues linked above to capture a few side paths from the conversation in this PR from before it closed.

PR #339, opened 2023-10-12
Feature/pyjson

### LINKED ISSUE #319: Add Critical Software decision point

### LINKED ISSUE #325: Create python package to facilitate json round-tripping

### LINKED ISSUE #330: Model CVSS v1, v2, v3 vector elements as decision points

ahouseholder:
Reopening to keep the PR active (#328 closed automatically when I was messing with the branch), but
(a) please don't push to this branch directly while it's in draft mode, and 
(b) there are a number of comments in the thread of #328 that should be separate issues entirely. 
This PR is about getting the python stuff in place. Content changes are out of scope for this PR.

- [x] Blocked by #321

- resolves #325 
- resolves #319 
- resolves #330


sei-vsarvepalli:
The relative path is what is from the root I think so 
https://github.com/ahouseholder/SSVC/blob/31de409580775cbc65cf1e5271efea9b436d09e1/src/test/test_schema.py#L12

should be more like "data/schema/Decision_Point.schema.json" like that. Also the ID field inside the schema needs to be updated to reflect the new location.  These two lines also need to be updated. 

https://github.com/ahouseholder/SSVC/blob/31de409580775cbc65cf1e5271efea9b436d09e1/data/schema/Decision_Point.schema.json#L4 

https://github.com/ahouseholder/SSVC/blob/31de409580775cbc65cf1e5271efea9b436d09e1/data/schema/Decision_Point_Group.schema.json#L4

ahouseholder:
I will be withdrawing this PR once I've chunked it up into a series of smaller PRs.

So far that includes:
- #346 
- #343 
- #342 

There will be a few more. I'll try to remember to mention this PR when I submit them.

ahouseholder:
Running tally: the following PRs, once merged, represent the entirety of what was originally in this PR.

- #346 
- #343 
- #342 
- #370

ahouseholder:
All of the subsequent fragments of this PR have been merged, so this one is no longer needed as a placeholder.

PR #341, opened 2023-10-16
Begin recording architecture decisions

ahouseholder:
This PR adds the `/docs/adr` directory and creates two starter architecture decisions
- 0000 to record architecture decisions
- 0001 to use MADR format for them

References:
- https://adr.github.io/
- https://adr.github.io/madr/

Rationale:
- We need a place to capture the outcomes of decisions within the project.
- ADR (and the MADR format in particular) seem to be a good enough place to start.

I can see this being where we capture things like
- why some things get versioned and others do not
- why SSVC decision points are ordered sets
- why SSVC outcome sets are ordered sets
- why we treat trees as examples not strict rules 
- and even some of the things that went into the original paper (like the decision trees vs other alternatives)


j--- (approved):
This is a good idea, thank you.

PR #350, opened 2023-10-17
add ADR proposals for decision point versioning.

ahouseholder:
Documenting two decisions coming out of Discussion #289

- Decision points are versioned using SemVer (i.e., _that_ they will be versioned)
- Specific details on how to apply SemVer to decision points (i.e., _how_ they will be versioned)

Note that the state of the branch in the PR assumes they are approved. If the discussion here concludes otherwise, we should update the status(es) accordingly before merging.


j--- (approved):
This works for me. Thanks for summarizing clearly.

Issue #335, opened 2023-10-09
Eisenhower matrix tutorial


ahouseholder:
As a simple tutorial on how to walk through the bootstrapping process, we could write a tutorial that models the Eisenhower matrix:

Importance: yes, no
Urgency: yes, no

Outcome set(s): 
 - do, schedule, delegate, delete 
 - do first, schedule, delegate, do not do

laurie-tyz:
I'm not fond of the "do not do".

do first       schedule              delegate            do not do
required     recommended      delegate           document and ignore
necessary   suggested           delegate             reject

Laurie

ahouseholder:
So just doing a Google image search for `eisenhower matrix`, I find:

| Important+Urgent | Important+Not Urgent | Not Important+Urgent | Not Important+Not Urgent|
|---|---|---|---|
| Do | Schedule| Delegate | Delete |
| Do Now | Schedule| Delegate | Delete |
| Do | Decide | Delegate | Delete |
| Do First and Reduce | Schedule | Delegate | Delete or Leave |
| Do | Schedule | Delegate | Limit |
| Do First | Delay | Delegate | Don't Do |
| Do | Schedule | Delegate | Eliminate |
| Reduce | Schedule | Delegate | Declutter |
| Do  | Schedule | Automate | Delete |
| Do | Plan | Delegate | Drop |

I'm going for very low translation overhead, which means using the words people already use in the way they use them.
My preference would be to not inject new words like "required", "necessary", "suggested", "reject" etc. because they're not part of the established vocabulary in how folks label the Eisenhower Matrix. 

So I guess it's "pick one from each column". _Do_, _Schedule_, and _Delegate_ seem straightforward enough. The question is what to call the not-important-not-urgent one. _Delete_ seems to be pretty common, but there is more variation in what people call this one.

ahouseholder:
Aside: It's not lost on me that RFC-2119's MUST, SHOULD, MAY, and MUST NOT are functionally equivalent. However those tend to show up in normative documents rather than in the kinds of operational decisions we're trying to model, so I don't know that we'd need them for anything. (Creating them as an outcome set is easy though if we chose to do so.)

j---:
[...]

PR #365, opened 2023-10-31
Policy Generator tool, first pass

### LINKED ISSUE #336: Tool to distribute outcome set across a decision point group
As a way to facilitate an SSVC bootstrapping exercise, we should create a tool that takes as input:

- a decision point group
- an outcome set

It should:

- map the decision point set into a partial order over its value combinations
- distribute the outcome set across those combinations
- perhaps with a few options
  - balanced: try to assign approximately the same number of leaf nodes to each outcome value
  - weighted: if the input includes weights on the outcome set, try to assign leaf nodes with approximately that frequency. 
  
If you think about it this way, you only have to implement the weighted version, then the balanced one is just a special default case where weight $w_i=1/N$ for an outcome set of size $|N|$

The approach I have in mind is to construct a digraph of the input space, do a topological sort, and then assign leaf nodes according to the weights.

This won't produce perfect policies, but it should get most users in the ballpark of something they could then tune for their own needs.

### LINKED ISSUE #337: MoSCoW outcome set
Model a standard MoSCoW outcome set:

- Must
- Should
- Could
- Won't

As with #335, the idea here is to give folks a vocabulary they might already be familiar with to help translate their prioritization methods to implementation in SSVC.

https://en.wikipedia.org/wiki/MoSCoW_method

ahouseholder:
This PR contains a python class that generates an SSVC decision policy (aka a "tree") that fully enumerates a set of decision points and their values, and assigns an outcome to each input value tuple. It does so by 

- constructing a partial order over all the input value tuples
- generating a topological sort over the resulting partial order
- assigning outcomes according to a set of provided weights
- emitting a CSV that replicates the format we've previously used to represent decision policies
- (one column for each decision point, plus an outcome column filled in with the assignments made)

If you run `policy_generator.py` directly, it outputs a policy that's rather close to the Deployer 2.1 tree.

At this point, my goal is to get the module in place with this PR. 

We will need to figure out what the UI is for it, and how to make it easier to use. Right now, you have to write python code to use it, an example of which is provided in `ssvc.policy_generator.main`, but it can probably be made easier to use. One usability issue I can see coming soon is how someone figures out which decision point (and version) to use as input, and how they represent their outcome set. My preference would be to create a separate issue to come back to the CLI side of this tool in the future.  For now, I think it's usable enough for someone comfortable with some light coding-by-example, such as a Jupyter notebook user. I created a separate issue to capture the UI followup task (#366) .

One notable addition in this PR is the creation of `src/ssvc/outcomes/base.py`, which contains a python representation of an `OutcomeGroup` and `OutcomeValues`, and `src/ssvc/outcomes/groups.py` which provides a variety of examples thereof. I think we discussed _not_ versioning outcome sets (I called them _outcome groups_ which have _outcome values_ here but if there's a preference to rename them simply _outcomes_ with _outcome values_, that's an easy change to make.)  

This PR
- resolves #336 
- resolves #337 
- partially addresses (by modeling the outcome set only) #335

I included #337 and #335 because they were good exercise for building `OutcomeGroup` objects.

### DIFF
--- a/src/ssvc/policy_generator.py
+++ b/src/ssvc/policy_generator.py
@@ -1,6 +1,7 @@
 #!/usr/bin/env python
 """
-Provides a Policy Generator class for SSVC decision point groups
+Provides a Policy Generator class for SSVC decision point groups.
+
 """
 #  Copyright (c) 2023 Carnegie Mellon University and Contributors.
 #  - see Contributors.md for a full list of Contributors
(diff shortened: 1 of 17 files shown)


ahouseholder:
This PR does not fully resolve, but makes a start at modeling the outcome sets that would support

- #337
- #335 

I included those because they were good exercise for building `OutcomeGroup` objects.

src/ssvc/policy_generator.py
j---, line 1: I tried to read this file carefully, and it looks good and makes sense to me.

j--- (approved):
This looks like it has all the parts it needs. I didn't do independent runtime testing.

src/ssvc/policy_generator.py
ahouseholder: I went ahead and merged it. If we run into bugs we can do separate issues later.

PR #368, opened 2023-10-31
Add Decision Point Group Versioning ADRs

### LINKED ISSUE #355: Create ADR(s) for Decision Point Group versioning
There is a whole discussion in Discussion #303 that needs to be detangled into one or more likely 2 ADRs, as in #350:

1. That Decision Point Groups are to be versioned using SemVer
2. How to apply SemVer to this problem (a specific rule set)

Once a PR that closes this issue has been merged, then Discussion #303 can also be closed.

ahouseholder:
This PR adds two ADRs:

- 0004 - SSVC Decision Point Groups are Versioned using SemVer
- 0005 - SSVC Decision Point Group Versioning Rules

As with #350, I separated these out so that the decision _that they are versioned_ is distinct from the _versioning rules_.

It is my intent that existing decision point groups already present in the python code should be consistent with these rules. However, the rules were developed during the process of creating the python code, and written down afterwards. So any discrepancy is basically the result of an out-of-order operation. We should correct any such discrepancies we detect, of course, whether by modifying the code to match the rules, or perhaps adjusting the rules if there's a good reason to do so.

- resolves #355

### DIFF
--- a/docs/adr/0003-ssvc-decision-point-versioning-rules.md
+++ b/docs/adr/0003-ssvc-decision-point-versioning-rules.md
@@ -1,5 +1,5 @@
 ---
-status: accepted
+status: superseded by [ADR-0006](0006-ssvc-decision-point-versioning-rules.md)
 date: 2023-10-17 
 deciders: adh, jspring, vssarvepalli, latyzenhaus, cgyarbrough, ehatleback
 ---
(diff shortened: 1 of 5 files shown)


ahouseholder:
PR #369 would add ADR 0006 to supersede ADR 0003. I made #369 a PR against this one since it's technically a separate decision for us, but the ADR numbering depends on the assumption that this PR #328 is merged as well (with ADR 0004 and ADR 0005). If for some reason those assumptions go unmet, then PR #369 should renumber its new ADR accordingly.

j--- (approved):

PR #371, opened 2023-11-02
Add build steps to python-app.yml

ahouseholder:
Now that we have a pyproject.toml, we can verify that the build process works too, and upload artifacts from each run

Artifact retention is set to 14 days


j--- (approved):
This looks like it should work.

PR #376, opened 2023-11-07
Add CWE-PoC list file

### LINKED ISSUE #158: List of CWE for "state of exploitation" value PoC "the vulnerability has a well-known method of exploitation."
Create a draft list of CWE-IDs where this condition described in the text holds. 
Goal is to get a first pass done, it will be hard to be exhaustive the list can always be refined later. 
Proposed format for the file is CSV with two fields
CWE-ID,URL of open-source project that captures the well-known method of exploitation

For example, if it is on-path attacker TLS certificate interception,
CWE-ID-300,http://www.squid-cache.org/Versions/

Reading the CWE documentation on this raises some issues.
https://cwe.mitre.org/data/definitions/300.html

CWE-ID-300 is about all on-path attacker scenarios, not just TLS. 
This is partly good -- Squid can also proxy HTTP or FTP, for example. If CWE-ID-300 also applies to email client/server authentication, then squid isn't sufficient evidence to cover that. NGINX does seem to cover that, so it isn't a problem as far as listing the CWE as meeting the requirements for the "state of exploitation" criteria. But we might want to add another line

CWE-ID-300,https://www.nginx.com/resources/wiki/start/topics/examples/imapproxyexample/

I'm happy for us to have multiple lines of evidence supporting one CWE-ID. The thing we really need to flag is example CVE-IDs that match the CWE-ID but do not match the intended "state of exploitation" criteria of "well-known method of exploitation." I can't think of one for on-path attacker. If there end up being CWE-IDs where the answer to the "state of exploitation" criteria here is "sometimes" then we will discuss how to handle that then. For now, I want to see how much value we can add by create the list of "definitely yes" CWEs for the question "well-known method of exploitation" and see if that is useful, and we can go from there.

koscinv:
This CSV file contains a list of CWEs which may result in a _PoC_ value for "state of exploitation" because "the vulnerability has a well-known method of exploitation." It contains links to some exploit tools for those CWEs that indicate a potential PoC. It also contains CWEs which could not be PoCs as well as reasoning behind them. All CWEs in this list have been found to map to CVEs.

Potentially closes #158


j--- (approved):
This looks good to me. 

I expect there are places where folks could quibble with whether a tool is "easy enough" to use to exploit any instance of one of the CWEs listed. But I think this is a considered and well-formed place for us to have detailed conversations about that question if anyone would like to raise those questions in the details. In some sense, if anyone has questions about how to map data for *State of Exploitation* at that level of detail, I'll be quite pleased. 

Thanks very much for putting this together!

data/csvs/cwe/possible-cwe-with-poc-examples.csv
ahouseholder, line 1: I moved this into a folder just for housekeeping. No changes to the csv from @koscinv

docs/reference/decision_points/exploitation.md
ahouseholder, line 1: In the interest of avoiding a follow-on issue to update the docs, I went ahead and added the CSV to the exploitation page and did a light revision to the page to make it fit in.

ahouseholder (approved):
This is a great addition, thanks for the help!

PR #377, opened 2023-11-08
Add CVSSv4 Decision Points

### LINKED ISSUE #345: Model CVSS v4 vector elements as decision points
Extends #330 for CVSS v4. Similar implementation to what is in #343.

### LINKED ISSUE #348: Map CVSSv4 equivalence sets to CVSSv4 decision points
CVSS v4 adds the concept of "equivalence sets", which are rules for converting CVSS v4 vector elements to categories in 6 different dimensions. We can easily model these as: DecisionPointGroup(CVSSv4 dp1, ..., etc.). We should do that and provide it to assist folks who are trying to map a CVSS-based process onto an SSVC-oriented decision model.

ahouseholder:
Issue #330  and PR #343 were a first pass at modeling CVSS v1, v2, and v3 as SSVC decision points and groups.

This PR extends that work to include CVSS v4.

Along the way, I uncovered a few issues with how the changes in #343 didn't quite line up right, so I've fixed those issues as well. 

## Summarizing changes:

- Add decision points for all CVSS v4 BTE vector elements
- Add decision points for CVSS v4 equivalence sets
- Individual CVSS decision point versions should be aligned with our SSVC DP versioning rules
- I simplified the various "Not Defined" values into two distinct versions: one with the key `ND` for CVSS v2 and one with key `X` for CVSS v3 and v4
- Added/updated helper methods and unit tests for generating the "modified" elements in v3 and v4
- Added a `dp_diff` method that can compare sequential versions of a decision point against most of our versioning rules and suggest whether it's a major, minor, or patch version change. If you run any of the individual decision point `.py` files you'll see what it does.
- Cleaned up a bunch of long string lines (wrapped them)
- Removed some default IDE cruft from various `__init__.py` files
- consolidates v1, v2, v3, and v4 decision point groups into a single module `ssvc.dp_groups.cvss.collections`
- add DP groups for Base, Base + Temporal, Base + Temporal + Environmental for older versions (following example of B, BT, BTE in v4)
- added copyright blob to modified files.

## Resolutions

- resolves #345 
- resolves #348 

## Out of scope

I am considering the following to be _out of scope_ for this PR:
- #393
- #392 (also depends on PR #365)

It's not that either of these are difficult, they are just distinct tasks that can be broken out separately.

## Remaining to do before taking this out of draft

Initially creating as draft PR because the following work is outstanding:

- [x] CVSS v4 supplemental metrics are not yet included. There is conceptual overlap with some existing SSVC decision points (Automatable, Value Density, Safety) that we'll need to consider. Probably not a problem, just noting it still needs to done.


j---:
FWIW, CVSS supplemental metric "Safety" "Yes/no" is conceptually a map to "public safety impact" in SSVC. This is identically true if we map the SSVC safety impact descriptions back to IEC/ISO 61508 explicitly. SSVC currently implicitly maps to it since SSVC uses FAA and CDC definitions which are based on 61508, whereas CVSS explicitly uses 61508.

j--- (approved):
All looks good to me. Thanks!

j---:
I thought this went out of draft, but now I'm not sure, as it looks like it's still in draft.

ahouseholder:
I was holding it in draft until I got the supplementals in place. That's now done.

Follow-up tasks after this merges:
- #389 
- #392
- #393

ahouseholder:
> FWIW, CVSS supplemental metric "Safety" "Yes/no" is conceptually a map to "public safety impact" in SSVC. This is identically true if we map the SSVC safety impact descriptions back to IEC/ISO 61508 explicitly. SSVC currently implicitly maps to it since SSVC uses FAA and CDC definitions which are based on 61508, whereas CVSS explicitly uses 61508.

To be addressed by
- #389

ahouseholder:
Note that #391 was also spawned out of the work leading to this PR.

PR #399, opened 2024-01-17
Add grid to homepage

ahouseholder:
Currently staged at https://certcc.github.io/SSVC-staging/

This PR adds:

- intro paragraph on site home page
- a 4x4 grid to direct folks from the home page to the various sections of the site
- upgrades mkdocs-material to the latest release
- fixes a deprecation warning about obsolete mkdocs extensions

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -130,8 +130,8 @@ markdown_extensions:
   - pymdownx.caret
   - pymdownx.details
   - pymdownx.emoji:
-      emoji_index: !!python/name:materialx.emoji.twemoji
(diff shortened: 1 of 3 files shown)


mkdocs.yml
ahouseholder, line 1: These two lines were causing a deprecation warning when mkdocs would run because the extension name has changed. This change fixes that warning.

requirements.txt
ahouseholder, line 1: Updated mkdocs-material requirement to use the current version. This upgrade is necessary to use the grid motif on the home page because it was recently added to the mkdocs-material release.

j--- (approved):
Looks good to me. Thanks!

Issue #360, opened 2023-10-19
Set up dependabot for version updates


ahouseholder:
https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuring-dependabot-version-updates

PR #400, opened 2024-01-23
Pin versions in requirements.txt

ahouseholder:
This is in preparation for enabling dependabot (#360). Once we turn on dependabot, it will start suggesting updates to this list.


SEI-gstrom (approved):

sei-vsarvepalli (approved):
Good to go! Allen - thanks.

PR #425, opened 2024-02-01
Add OutcomeGroup for TheParanoids PrioritizedRiskRemediation

### LINKED ISSUE #358: Model theparanoids outcome set
Based on https://github.com/theparanoids/PrioritizedRiskRemediation#risk-based-decision-tree-decisions

There they define a 6-level outcome set (I'm listing them in "least" to "most" order, which is the opposite of their number scale)

| Value | Description |
|:---:|---|
| 5 | Track |
| 4 | Track Closely |
| 3 | Attend |
| 2 | Attend |
| 1 | Act |
| 0 | Act ASAP |

I'm not suggesting that we directly adopt these values or names. It appears they adapted the names from the CISA ones (see #356). I'm not a fan of this name set, for what its worth.

However, I suppose it's more about providing a higher granularity 6-level outcome set. I've come across at least one other service management process that has six priority levels, so it seems like something we should be providing as one of our "standard bricks".

ahouseholder:
- resolves #358

### DIFF
--- a/src/ssvc/outcomes/groups.py
+++ b/src/ssvc/outcomes/groups.py
@@ -133,12 +133,12 @@ THE_PARANOIDS = OutcomeGroup(
     name="theParanoids",
     description="PrioritizedRiskRemediation outcome group based on TheParanoids.",
     outcomes=(
-        OutcomeValue(name="Track", key="5", description="Track"),
-        OutcomeValue(name="Track Closely", key="4", description="Track Closely"),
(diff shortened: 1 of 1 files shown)


j--- (approved):
It bugs me that 2 and 3 have the same label, but that is a question / issue for the source material, not our representation of it, which (I agree) should represent it as is.

ahouseholder:
I added numbers to the outcome names in c1f383a to distinguish the two _Attend_ values for now. We can revisit this in the future if it turns out to be too awkward.

PR #427, opened 2024-02-01
Add CISA custom decision points

### LINKED ISSUE #357: Add decision points based on CISA SSVC
Based on https://www.cisa.gov/sites/default/files/publications/cisa-ssvc-guide%20508c.pdf

I'm not going to duplicate a few pages of that doc in this issue. But this issue is recommending that we consider modeling the following items defined in that doc as decision points:

| Decision Point | Values |
|:---:|---|
| Public Well-Being Impact | Minimal, Material, Irreversible |
| Mission Prevalence | Minimal, Support, Essential |
 
These then get rolled into a combo named _Mission and Well-Being Impact_, which we could also model:

| Decision Point | Values |
|:---:|---|
| Mission and Well-Being Impact | Low, Medium, High |

These are similar to existing SSVC decision points `Safety Impact`, `Mission Impact`, and `Human Impact`, but seem distinct enough that maybe they should be represented separately. Or, maybe I'm misremembering history but they might actually reflect _earlier_ versions of those things and so we should maybe adjust the version numbers on our existing ones to reflect that and add these as whatever the current version is -1.

So those are open questions:

- [ ] Is `Public Well-Being Impact` an earlier version of `Public Safety Impact` or is it just something distinct?
- [ ] Is `Mission Prevalence` an earlier version of `Mission Impact` or is it something distinct?
- [ ] Is `Mission and Well-Being Impact` an earlier version of `Human Impact` or something distinct?

Even if they turn out to be just older versions of the same concepts, we should probably still model them to reflect the evolution path for when anyone using CISA's decision points wants to upgrade to newer decision points representing similar concepts.

---

There is a section on _Mitigation Status_ that I do not understand, as it's not represented in the decision tree later in the doc, nor does it resolve into a single dimension or combination as with _Mission and Well-Being Impact_. Nevertheless, for recording purposes, it appears to have three substates each with two values:

| Decision Point | Values |
|:---:|---|
| Minimal | Available, Unavailable |
| System change difficulty | Low, High |
| Type | Fix, Workaround |

It's certainly trivial to copy and paste words out of that doc into the data structures to represent these last three as decision points, but we'd have to rename them to have clearer meaning (E.g., _Minimal_ is not a dimension, and _Type_ would have to at least be "Mitigation Type"). Regardless, I don't see how these three items are rolled up into something representing _Mitigation Status_ or what its possible values are.

Perhaps we could get clarification from CISA on this.

- [ ] How are `Minimal`, `System Change Difficulty`, and `Type` reflected in a decision?
- [ ] Is there a roll-up `Mitigation Status` decision point that somehow combines these? If so, how?

---
 
My reading is that `Technical Impact` and `Automatable` are used consistently with what we already have.

---
Related issues
- #352 covers renaming one of the `Exploitation` values which will make it match what CISA is using
- #356 covers the CISA outcome set

ahouseholder:
- resolves #357

This PR adds decision points:

- mission prevalence
- public well-being impact
- mission and well-being impact

Note that while the PR also updates the generated `.json` and `.md` files, these files are not linked within the documentation site as we haven't modeled the policy yet. So these objects are available for use in the python code but not otherwise integrated into the docs.


docs/_generated/decision_points/mission_and_well-being_impact_1_0_0.md
```diff
@@ -0,0 +1,18 @@
+<!-- This content is autogenerated by doctools.py. Do not Edit. -->
+!!! note "Mission and Well-Being Impact v1.0.0"
+
+    === "Text" 
+    
+        Mission and Well-Being Impact is a combination of Mission Impact and Public Well-Being impacts.
```
j---, line 6: Should say Mission Prevalence

j--- (changes requested):
Small typo in docs/_generated/decision_points/mission_and_well-being_impact_1_0_0.md

Otherwise, perfect.

docs/_generated/decision_points/mission_and_well-being_impact_1_0_0.md
ahouseholder: fixed in 509942b

j--- (approved):

Issue #353, opened 2023-10-17
Update description of `Exploitation:PoC`


ahouseholder:
Regarding

https://github.com/CERTCC/SSVC/blob/798ff57645f9331a46331d44f54422bdf6d95a71/src/ssvc/decision_points/exploitation.py#L16-L22

@j--- wrote:

> Do we actually think (1) and (2) are operationalizable? 
CISA's SSVC version of *State of Exploitation* removes them https://www.cisa.gov/sites/default/files/publications/cisa-ssvc-guide%20508c.pdf
Maybe we should move for consistency between instances of the decision point?

_Originally posted by @j--- in https://github.com/CERTCC/SSVC/pull/328#discussion_r1355999465_
            

Quoting the CISA doc referenced above:

> (State of) Exploitation
Evidence of Active Exploitation of a Vulnerability
This measure determines the present state of exploitation of the vulnerability. It does not predict future exploitation or
measure feasibility or ease of adversary development of future exploit code; rather, it acknowledges available
information at time of analysis. As the current state of exploitation often changes over time, answers should be timestamped. Sources that can provide public reporting of active exploitation include the vendor’s vulnerability
notification, the National Vulnerability Database (NVD) and links therein, bulletins from relevant information sharing
and analysis centers (ISACs), and reliable threat reports that list either the CVE-ID or common name of the
vulnerability.
Table 2: Exploitation Decision Values
Value Definition
None There is no evidence of active exploitation and no public proof of concept (PoC)
of how to exploit the vulnerability.
Public PoC One of the following is true: (1) Typical public PoC exists in sources such as
Metasploit or websites like ExploitDB; or (2) the vulnerability has a well-known
method of exploitation. Some examples of condition (2) are open-source web
proxies that serve as the PoC code for how to exploit any vulnerability in the vein
of improper validation of Transport Layer Security (TLS) certificates, and
Wireshark serving as a PoC for packet replay attacks on ethernet or Wi-Fi
networks.
Active Shared, observable, and reliable evidence that cyber threat actors have used the
exploit in the wild; the public reporting is from a credible source.

ahouseholder:
Note to future resolver (probably me, but you never know): If you're resolving this one and #352 remains unresolved, you should probably go ahead and pick that one up too.

ahouseholder:
Relevant side discussion that might affect this issue:

- #363

PR #442, opened 2024-02-07
Update Exploitation:PoC definition

ccullen-cert:
I changed the content to the following:

description="A POC can be accessed either on the clear web or in illicit locations, regardless of a paywall"
    " or the vulnerability has a well-known method of exploitation.",
    
I believe this covers the intended goal behind this decision point. There are an infinite amount of variations and combinations for PoC to be accessed by a TA, and paywalls won't stop a TA from gaining access to them. The point, from my understanding, is that this decision point is used to signify an incoming likelihood of exploitation by a threat actor. If the PoC can be accessed, even determined lower level TAs can acquire it from behind a paywall, even if it is provided by a legitimate enterprise.  
There are other nuances to the decision point, but that was what guided by decision in changing this up. 

The second point: "or the vulnerability has a well-known method of exploitation" covers an instance where a PoC may not have an explicit PoC or otherwise formal place where individuals can acquire a PoC, but the vulnerability can be exploited through already known methods that do not require much effort to accomplish or put together.


src/ssvc/decision_points/exploitation.py
```diff
@@ -14,11 +14,10 @@
 )
 
 POC = SsvcDecisionPointValue(
-    name="PoC",
+    name="Public PoC",
     key="P",
-    description="One of the following cases is true: (1) private evidence of exploitation is attested but not shared; "
-    "(2) widespread hearsay attests to exploitation; (3) typical public PoC in places such as Metasploit"
-    " or ExploitDB; or (4) the vulnerability has a well-known method of exploitation.",
+    description="A POC can be accessed either on the clear web or in illicit locations, regardless of a paywall"
```
j---, line 19: Can we make this consistent with the CISA description rather than trying to spin something new?  Per #353 
CISA's SSVC version of State of Exploitation removes them https://www.cisa.gov/sites/default/files/publications/cisa-ssvc-guide%20508c.pdf

Also, well-known methods of exploitation have been ground out a bit in #376, so we could productively reference that in the documentation.

j--- (changes requested):
In addition to text comments, I think that https://github.com/CERTCC/SSVC/blob/main/src/ssvc/decision_points/exploitation.py is the canonical place that the text needs to be updated, so after we agree on text let's make sure it's reflected there also before PR is finalized. 
Thanks!

ahouseholder:
If the wording changes in the python, then we may also need to bump a version number. I'll have to take a look at that after the other changes settle out but before we merge this one.

src/ssvc/decision_points/exploitation.py
ahouseholder: re: #376, it's already part of the Exploitation doc page. 

See https://certcc.github.io/SSVC-staging/reference/decision_points/exploitation/

src/ssvc/decision_points/exploitation.py
```diff
@@ -14,11 +14,10 @@
 )
 
 POC = SsvcDecisionPointValue(
```
ahouseholder, line 16: This should probably get a new `POC_2` name and a new version of the decision point rolled in.

src/ssvc/decision_points/exploitation.py
ahouseholder: So taking the words from the CISA doc, edited to stay consistent with [ADR-0007 Object Descriptions Exclude Examples
](https://github.com/CERTCC/SSVC/blob/main/docs/adr/0007-descriptions-exclude-examples.md), it seems like we might want to use the following text:

| Value | Description |
| --- | --- |
| None | There is no evidence of active exploitation and no public proof of concept (PoC) of how to exploit the vulnerability. |
| Public PoC | One of the following is true: (1) Typical public PoC exists in sources such as Metasploit or websites like ExploitDB; or (2) the vulnerability has a well-known method of exploitation.
| Active | Shared, observable, and reliable evidence that cyber threat actors have used the exploit in the wild; the public reporting is from a credible source. |

src/ssvc/decision_points/exploitation.py
j---: I think the version number needs to change here, right? 
https://github.com/CERTCC/SSVC/blob/524ad63e76dec34f1bed25a93a862e45feac317b/src/ssvc/decision_points/exploitation.py#L39

And what a new version means is that `exploitation` uses `POC` instead of something else. So if you leave this one as `POC`, then the legacy version of `POC` needs a new name so that `exploitation` version 1.1.0 can be distinguished from version 1.0.0 (current version). 

I think it'll be 1.1.0, at least.

src/ssvc/decision_points/exploitation.py
j---: Thanks!

src/ssvc/decision_points/exploitation.py
ahouseholder: yes, it needs to be a new version. I just

- restored the original POC definition in ad99a1d 
- re-added the new one as POC_2 in 5604ee6
- updated copyright/docstring (my local tooling does this automatically) in 99b7125
- regenerated examples from the new version in 9a890aa

I'm going to mark this conversation resolved and re-request review on it.

src/ssvc/decision_points/exploitation.py
```diff
@@ -14,11 +14,14 @@
...
-    name="Public PoC",
+    name="PoC",
```
ahouseholder, line 29: I reversed these changes. Because we're versioning decision points, we can't modify older ones in place, we need to append new ones. That gets fixed in the next commit.

src/ssvc/decision_points/exploitation.py
```diff
@@ -13,14 +14,19 @@
...
     description="One of the following cases is true: (1) private evidence of exploitation is attested but not shared; "
     "(2) widespread hearsay attests to exploitation; (3) typical public PoC in places such as Metasploit"
     " or ExploitDB; or (4) the vulnerability has a well-known method of exploitation.",
 )
 
+POC_2 = SsvcDecisionPointValue(
+    name="Public PoC",
+    key="P",
+    description="One of the following is true: (1) Typical public PoC exists in sources such as Metasploit or websites like ExploitDB; or (2) the vulnerability has a well-known method of exploitation.",
+)
```
ahouseholder, line 40: This is how to modify an existing decision point value: copy it to a new variable name, then create a new decision point (below) that uses it. Modifying existing decision point values can't happen or else we lose versioning.

src/ssvc/decision_points/exploitation.py
```diff
@@ -40,14 +46,28 @@ def _strip_spaces(s):
...
+
+EXPLOITATION_2 = SsvcDecisionPoint(
+    name="Exploitation",
+    description="The present state of exploitation of the vulnerability.",
+    key="E",
+    version="1.1.0",
+    values=(
+        EXP_NONE,
+        POC_2,
         ACTIVE,
```
ahouseholder, line 73: Here I added a new version of the Exploitation decision point that picks up the new POC_2 value.

Just noticing I should have named it EXPLOITATION_1_1_0. I'll fix that in another commit.

src/ssvc/decision_points/exploitation.py
```diff
@@ -40,14 +46,28 @@ def _strip_spaces(s):
...
     ),
 )
 
 
 def main():
-    print(EXPLOITATION_1.to_json(indent=2))
+    versions = [EXPLOITATION_1, EXPLOITATION_2]
+
+    print_versions_and_diffs(versions)
```
ahouseholder, line 81: When you run this module as a standalone, it'll suggest whether it's a major, minor, or patch version change based on the delta. It suggested major version because the name changed from PoC to Public PoC, but I overrode that and made it a minor version change since the two values are semantically compatible.

ahouseholder:
Based on this PR, I also created the following how-to in the SSVC wiki for future reference:

https://github.com/CERTCC/SSVC/wiki/Modifying-Decision-Point-Values

Thanks especially to @ccullen-cert for being an unwitting guinea pig being the first person other than me to attempt modifying a decision point definition in the python files, so it wasn't until now that I realized how much wasn't obvious about the versioning process. Hence the doc linked above. Hopefully that makes it easier the next time through.

ahouseholder (approved):
I'm good to go with this as it currently stands.

sei-vsarvepalli (approved):
All looks clean to me.

ahouseholder:
@j--- since you had originally requested changes, I think you have to re-review and approve this to clear the way to merge.

j--- (approved):
Thanks for sorting this all out!

PR #443, opened 2024-02-07
Fix broken links

### LINKED ISSUE #430: Fix the "#links" throughout
Because of the former markdown-to-latex-to-pdf process and the conversion to separate markdown files as distinct web pages, we now have a lot of links that appear to be to sections of the page they're on, but those sections don't exist. They should in fact be links to pages, but we'll need to figure out for each of them what page they need to point to.

I'm jotting this down as a quick issue without much accompanying detail so we don't forget about it. But a rough `find | xargs grep` check has like 500+ such strings in the markdown files.

ahouseholder:
- resolves #430 
 
This PR is an attempt to fix as many broken links as possible following the conversion from our PDF-oriented workflow to our mkdocs-oriented workflow.

Converting from one massive file to a website left us with a lot of section links (links to heading anchors like `[Utility](#utility)` that need to be replaced with links to files `[Utility](../reference/decision_points/utility.md)`.

By a quick search of the markdown content, I think there were over 500 such links to start. I believe I've fixed them all here.

Output from the python `linkchecker` tool:

```
Statistics:
Downloaded: 5.43MB.
Content types: 6 image, 94 text, 0 video, 0 audio, 10 application, 2 mail and 250 other.
URL lengths: min=15, max=899, avg=84.

That's it. 362 links in 362 URLs checked. 0 warnings found. 0 errors found.
Stopped checking at 2024-02-14 12:07:05-004 (50 seconds)
```

Once merged, this PR puts us into a "known-good" ground state for link checking as well. I submitted a separate PR #465  as a downstream dependency on this one that will add a `linkchecker` acceptance test workflow for all future PRs. 

- prerequisite for #465

**Note** Once approved, this PR should be _squash merged_ rather than a normal merge due to the fact that the individual commits have almost no meaning as they're all intermediate work-in-progress fixes.


docs/howto/bootstrap/collect.md
```diff
@@ -43,4 +43,87 @@ flowchart LR
      They also write a script that parses the data from the threat feeds and applies the data map to assign a value to 
      the _Exploitation_ decision point.
 
-{% include-markdown '../evidence_gathering.md' heading-offset=1 %}
```
ahouseholder, line 46: As I was fixing links, I also took advantage of some opportunities to eliminate the need for some file includes by just merging the included content into the parent file.

docs/howto/bootstrap/use.md
```diff
@@ -83,4 +83,154 @@ flowchart LR
     SSVC does not prescribe any particular response process, but it does provide a structured way to make decisions 
     within the response process.
 
-{% include-markdown '../communicating_results.md' heading-offset=1 %}
```
ahouseholder, line 86: As I was fixing links, I also took advantage of some opportunities to eliminate the need for some file includes by just merging the included content into the parent file.

docs/howto/communicating_results.md
ahouseholder, line 1: As I was fixing links, I also took advantage of some opportunities to eliminate the need for some file includes by just merging the included content into the parent file.

ahouseholder (commented):
As I was fixing links, I also took advantage of some opportunities to eliminate the need for some file includes by just merging the included content into the parent file. Hence there are a few file changes that look like they're adding content to files (which technically they are), but those are offset by removing the files that originally included that content.

j--- (approved):
thanks so much for fixing all of these.

PR #446, opened 2024-02-09
merge report credibility description with its decision point page

ahouseholder:
This is a small PR that recombines two pages that probably should have stayed together in the first place.

Before: separate pages describe report credibility heuristics and the report credibility decision point
After: all that content is now part of the report credibility decision point page

Changes to `topics/index.md` and `mkdocs.yml` were incidental to making this change.


j--- (approved):
Makes sense to me!

PR #451, opened 2024-02-09
Remove abbreviated format from docs

### LINKED ISSUE #333: Reduce emphasis or possibly deprecate the vector format
Since we are moving toward decision points being the primary data objects, this began as a discussion over whether the "vector format" should include data point version information.

On further discussion, the consensus appeared to be that the usefulness of the vector format is in doubt, and the main argument in favor of it was that it resembles CVSS vector representation. But that didn't feel very compelling to the conversants at the time. 

So the direction we came up with was to de-emphasize the "vector format" in favor of the `.json` representation, and encourage people to use words rather than shorthand notation to talk about SSVC decision points and values. 

Resolving this issue will involve either choosing to deprecate the vector format entirely, or just de-emphasize it in the docs (perhaps it survives as an example of how to use decision point keys?)

This issue is the driver behind Issue #332

ahouseholder:
This PR removes the "abbreviated format" aka "vector format" from the documentation.

- ran into this while working on #443
- resolves #333

### DIFF
--- a/docs/howto/communicating_results.md
+++ b/docs/howto/communicating_results.md
@@ -14,59 +14,9 @@ This section discusses the situation where one stakeholder, usually a supplier o
 
 ## Communication Formats
 
-We recommend two structured communication formats, abbreviated and full.
-The goal of the abbreviated format is to fill a need for providing identifying information about a vulnerability or decision in charts, graphs, and tables. Therefore, the abbreviated format is not designed to stand alone.
-The goal of the full format is to capture all the context and details about a decision or work item in a clear and machine-readable way.
+We recommend a JSON format for communicating SSVC information about a specific vulnerability.
+The goal of this format is to capture all the context and details about a decision or work item in a clear and machine-readable way.
 
-### Abbreviated Format
-
-SSVC abbreviated form borrows directly from the CVSS “vector string” notation.
-Since the purpose of the abbreviated form is to provide labels for charts and graphics, it does not stand alone.
-The basic format for SSVC is:
-```
-SSVC/(version)/(decision point):(value)[/decision point:value[/decision point:value[...]]][/time]/
-```
-Where `version` is `v2` if it is based on this current version of the SSVC.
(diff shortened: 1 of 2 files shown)


docs/howto/communicating_results.md
```diff
@@ -90,17 +40,14 @@ A timely partial warning is better than a complete warning that is too late.
 The basic guidance is that the analyst should communicate all of the vulnerability's possible states, to the best of the analyst's knowledge.
 If the analyst knows nothing, all states are possible.
 For example, [*Utility*](#utility) may be [laborious](#utility), [efficient](#utility), or [super effective](#utility).
-In abbreviated form, write this as `U:LESe`.
-Since a capital letter always indicates a new value, this is unambiguous.
 
-The reason a stakeholder might publish something like `U:LESe` is that it expresses that the analyst thought about [*Utility*](#utility) but does not have anything to communicate.
+The reason a stakeholder might publish a decision point with more than one value is that it expresses that the analyst thought about [*Utility*](#utility) but does not have anything to communicate.
```
j---, line 44: "more than one value" is not the same as "all values". I think the sentence only makes sense in the context of "all values" ? 
Also possible this sentence is just confusing / not something we would recommend people do, so possible option is to remove the whole discussion of the topic, which more broadly is "how to document an analyst looked but didn't learn anything"

That would probably be a separate issue? So the minimum required to just address this translation here is s/more than one value/all values/

j--- (changes requested):
Not sure on the best way forward to address the one comment.

docs/howto/communicating_results.md
ahouseholder: took the easy way out in e01f6a5

I'll spawn a separate issue to revisit the _Partial or Incomplete Information_ section in the future

j--- (approved):

ahouseholder:
Leaving a breadcrumb here that there is some additional discussion in #333 about the consequences of this PR.

PR #464, opened 2024-02-14
Bump site copyright

### LINKED ISSUE #460: Update copyright in footer

ccullen-cert:
Updated MKDocs with the new copyright date of 2024

- resolves #460


ahouseholder:
I think it's because you based your change from a branch already in your fork rather than from the https://github.com/CERTCC/SSVC/tree/main project branch. Those other commits were the history that was already merged in #442.

Because the net change is just the one line in mkdocs.yml it's fine for this PR, but we should sort out how to avoid this in the future. (I've seen other PRs do this too, not sure what the root cause is though.)

ahouseholder (approved):

PR #465, opened 2024-02-14
Create link_checker.yml

### LINKED ISSUE #404: Add link checker workflow as PR acceptance test
As we shift toward a web-based documentation project, we need to add a link checker acceptance test for PRs. This should probably be set to run on PRs against `main` as well as on pushes to the `main` and `publish` branches.

I have not looked for link checker actions in GH marketplace, but there's a good chance one already exists and we just need to define a workflow to use it.

ahouseholder:
Adds a github workflow that builds the site and checks links anytime a PR wants to change a markdown file.

- resolves #404 

I feel obligated to acknowledge the entirely unintentional fact that issue #404 is the absolute best issue number possible for that specific issue.

This PR is a direct follow-on to 

- #443 

and should be merged immediately after #443 is merged. Rationale: #443 puts the content in a known-good state with no broken links that `linkchecker` can find. So if we add this workflow immediately afterwards, we can ensure that we keep it that way going forward.  Merging this PR _before_ #443 would just leave us with a failing workflow because there are a lot of broken links in the markdown up to that point.

### DIFF
--- a/.github/workflows/link_checker.yml
+++ b/.github/workflows/link_checker.yml
@@ -1,0 +1,41 @@
+name: Link Checker
+on:
+  pull_request:
+    paths:
+      # run on any PR that modifies a markdown file
+      - '**/*.md'
+      # run on any PR that changes this workflow
+      - .github/workflows/linkchecker.yml
+  # let us trigger it manually
+  workflow_dispatch:
+
+jobs:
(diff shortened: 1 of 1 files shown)


j--- (approved):
I'm so glad this will solve issue 404.

PR #470, opened 2024-02-15
Fix how we load csv files

ahouseholder:
Merging in #465 then manually running the link checker uncovered a bug in how we had been loading the CSV files in a few places. This PR fixes that.


ahouseholder:
I intentionally bypassed the approval process on this one since it was a bug that only surfaced after #465 was merged and the linkchecker workflow failed when manually run. With this PR, the linkchecker workflow ran automatically and succeeded on this branch before merging, so I'm confident that the problem was resolved.

PR #473, opened 2024-02-15
Add topo sort to csv analyzer and policy generator

ahouseholder:
This PR 
- adds some topological sort checks to the csv analyzer and policy generator scripts.
- reorders the `report public` and `report credibility` decision points to match expected directionality of values

These features were added after I noticed that report public and report credibility were improperly ordered according to the policy generator assumption that decision points are ordered from lowest-to-highest along a dimension that correlates to "likely to act".

### DIFF
--- a/data/json/decision_points/report_credibility_1_0_0.json
+++ b/data/json/decision_points/report_credibility_1_0_0.json
@@ -6,14 +6,14 @@
   "description": "Is the report credible?",
   "values": [
     {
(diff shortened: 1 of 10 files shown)


cgyarbrough (commented):
While I'm not a coder, this seems to be almost an administrative change. The concept and my uninformed view of the code seems to be ok, but I'll defer to either Vijay or Jono to approve the changes.

PR #512, opened 2024-02-28
Ccullen cert patch 2

### LINKED ISSUE #381: Bootstrapping section should mention stakeholder roles involved in each phase
This came out of a conversation that was generally around the question "Who needs to be involved in each step of the bootstrapping process?" and an idea that came up was that Decision Point concepts and values should generally be a boundary object understandable to both technical management and analysts, whereas datamapping is more about the analyst-level perspective that says how data will be used as evidence for assigning decision point values.

So the idea for this issue is to revise the bootstrapping docs to reflect those role distinctions.

### LINKED ISSUE #493: Fix TODO in `items_with_same_priority.md`
Fix the line

> `However, as discussed {== previously (see page 4) ==}, CVSS is designed to be accurate only within +/- 0.5,`

to point to the appropriate location, (or revise text to make it unnecessary).

Figuring out the correct reference probably requires going back to the PDF.

ccullen-cert:
This request is to address https://github.com/CERTCC/SSVC/issues/493 and https://github.com/CERTCC/SSVC/issues/381.
Its showing I modified index.md but the change is the same. 

- resolves #493 
- resolves #381

### DIFF
--- a/docs/howto/bootstrap/prepare.md
+++ b/docs/howto/bootstrap/prepare.md
@@ -3,11 +3,21 @@
 Preparing to use SSVC involves defining a decision you want to make,
 the information you need to make that decision, and the policy you want to use to make that decision.
 
-During the bootstrapping process, internal discussions regarding what segments of an organization are to handle the decision making process need to be defined
-prior to engagement. Risk owners must be involved in the development of the risk management policy represented by SSVC. ITSM (Information Technology Services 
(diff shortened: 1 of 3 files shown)


ahouseholder (approved):
I'm approving this since the net change is ok. 

The change history on this branch is a little weird because it looks like it started as a branch off of your own fork's main which was already out of sync with CERTCC/SSVC:main, but that's going to disappear as soon as we squash merge this. Not a big deal for this one, but starting future branches from CERTCC/SSVC:main might make it easier to work with.

Issue #438, opened 2024-02-05
UI suggestions for SSVC-calc


ahouseholder:
Capturing feedback from @bkoo in #412:

> I mostly echo what @ehatleback has already mentioned with an emphasis on his second point. By having the selections always visible to enable visual continuity, it also will help see the final decision point as both modes (Graphic/Analyst) block the final result with an export window. At the very least, perhaps move the export window out of the way so that the result can be seen? 
<img width="1132" alt="Screenshot 2024-01-31 at 5 48 00 PM" src="https://github.com/CERTCC/SSVC/assets/111892109/b0cde54d-4737-48f3-beb1-f7679925d8ea">
<img width="1139" alt="Screenshot 2024-01-31 at 5 48 20 PM" src="https://github.com/CERTCC/SSVC/assets/111892109/8cb02b68-9e7d-4d90-9c92-17b9fe17d4a8">

_Originally posted by @sei-bkoo in https://github.com/CERTCC/SSVC/issues/412#issuecomment-1920124439_
            

I'm creating this as a separate issue from #412 since it covers new changes and #412 is just about making sure the existing calculator works in the new layout.

ahouseholder:
also capturing @ehatleback's [comments](https://github.com/CERTCC/SSVC/issues/412#issuecomment-1919764611) in #412:

> In the first decision point, scrolling over the options highlights the top and bottom options (None and Active) in red, but the middle option (PoC) does not include the red highlighting.
>
> Additionally, I'd suggest that the appearance of the options when I hover is somehow a bit jarring or disorienting. A better option might be for the available selections to either be available all the time on the screen (rather than appearing when you hover), or to appear when you hover, but always to appear in the same location on the screen, like at the bottom.
>
> Finally, the options that appear when you hover frequently look like they're a little bit too big for the window they appear in; the text is right at the very edge of, and sometimes even cut off a bit from, the box that contains the appearing text.

ahouseholder:
It is not clear to me whether this issue was fully or only partially resolved by 

- #513

Also note there are some more possibly related comments in #514. Both #514 and #438 should be revised into separately resolvable issues if possible.

sei-vsarvepalli:
Yep only partially resolved the Export window issue. The other appearance related ones remain.

sei-vsarvepalli:
I think this is resolved now. At least both these issues are solved now.

PR #513, opened 2024-02-29
SSVC Calculator minor updates

sei-vsarvepalli:
This is a simple first attempt to make the SSVC calculator embedded with the mkdocs so it confirms to the new website that will be built. There is not much that should change. The https://sei-vsarvepalli.github.io/SSVC/ssvc-calc/ should show pretty close to what the SSVC site will look like.

Few visual bugs resolved from #438 
1. The PoC vs poc labels in JSON causes the red cover box to not appear on the popup window
2. Export popup covers the current view of the data being exported.


ehatleback:
When I scroll over some text (for example, the "Automatable" text), the popup still cuts off the edge of the wording.  Additionally, the "jarring" feeling of so many hover-pop-ups remains; it really would be good to make (for example) the whole bottom of the screen a follow-along static presentation of the options and corresponding text at those tree locations (where the tree itself is displayed in the upper half of the screen).  As it stands - and maybe this is just my own individual experience - as I kept clicking around and working with it, and items were flashing up all over in different positions on the screen, I actually began to get a little bit of a headache.

Also, hovering over the dots / buttons gives the satisfying "hand" icon suggesting that I can click there, but hovering over the word for each choice that's aligned along the tree/branch line doesn't display that "hand" / "clickable" icon, even though I've verified that I can in fact click the word to make my choice.

ahouseholder (approved):
Looks good to me. All we need for this PR is for the existing calculator to have the site header and footer wrapped around it. 

UI/UX feedback can be spun off into a separate issue to be handled later.

Masato20240324:
The explanation for "none" in "Mission Impact" in Deployer.json is "Little to no impact up to degradation of non-essential functions; chronic degradation would eventually harm essential functions. (aka Non-Essential Degraded)" . This has the same description as "degraded". Sorry for not being able to provide an alternative English sentence.

sei-vsarvepalli:
> The explanation for "none" in "Mission Impact" in Deployer.json is "Little to no impact up to degradation of non-essential functions; chronic degradation would eventually harm essential functions. (aka Non-Essential Degraded)" . This has the same description as "degraded". Sorry for not being able to provide an alternative English sentence.

Hello Masato, 

Thank you for reaching out to us.

In the `Mission Impact` - `none` and `degraded` are basically the same, therefore the option `none` has been eliminated in the newer schema. It is present only to maintain some backward compatibility. Perhaps it is best to update all the schema to match newer descriptions?

PR #518, opened 2024-03-01
Catch up README.md to current state

### LINKED ISSUE #501: update README to reflect outcomes of 202403 project
Once we finish the project efforts update the main README (https://github.com/CERTCC/SSVC/blob/main/README.md) to reflect updates, including a very prominent link to the updated website.

ahouseholder:
- resolves #501 

This PR updates the repository README to reflect the current work status.

It also eliminates the now-obsoleted `/drafts/` directory that used to house the generated PDFs. The final PDF draft is now housed in `/pdfs/` along with the previous published PDF versions.

The `ssvc.html` draft that corresponded to the `ssvc.pdf` draft has been removed. 

# Do not merge prior to 7 March 2024

Or at least know that by doing so there will be broken links in `/README.md`. The linkchecker action is only checking the rendered static site, not the in-repository markdown files, which is why that test will pass even if there are bad links in the README.md. (If that's a problem we think we should resolve, we should create a new issue to address that in the future.)


j--- (approved):
Thanks. LGTM
Remember not to merge until Mar 7

sei-vsarvepalli (approved):
its time isn't it?

PR #527, opened 2024-03-05
Add js to fix latex rendering on page load

ahouseholder:
This PR alters how `mathjax.js` renders $\LaTeX$ in pages. This fixes a minor annoyance that pages sometimes fail to render math correctly on the first load, but look fine on reload. The change here eliminates the need to reload the page.

Copied from CERTCC/Vultron#74

### DIFF
--- a/docs/javascripts/mathjax.js
+++ b/docs/javascripts/mathjax.js
@@ -12,5 +12,8 @@ window.MathJax = {
 };
 
 document$.subscribe(() => {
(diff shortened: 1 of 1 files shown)


sei-vsarvepalli (approved):
Seems to fix the onload issue. May be long term better to avoid MathJax remaining a global var to avoid any client based rewrite/prototype abuse.

PR #528, opened 2024-03-05
draft of update for cvss v4

### LINKED ISSUE #284: Add text about how Technical Impact relates to CVSS v4
https://github.com/CERTCC/SSVC/blob/edb719047d5b5b974dfd23dfb0c5677388c6eefb/doc/md_src_files/12_related_systems.md?plain=1#L29

We don't need to delete the references to CVSS v3.1. That is still a very active standard and the discussion about it will be relevant for probably another 10 years. 

However, CVSS v4 removes scope and separates out Vulnerable System Impact and Subsequent System Impact. *Technical Impact* in SSVC can now be cleanly derived as a function of CVSS v4 Vulnerable System Impact, if desired. We may want to discuss what that function would be. I know we've discussed this, but I don't see it in https://github.com/CERTCC/SSVC/blob/main/doc/md_src_files/05_02_technical_impact.md or in the section referenced above. 

I think loosely what I'd suggest is if VC:H AND VI:H THEN _Technical Impact:Total_. ELSE _Technical Impact:Partial_ 
(VC = Vulnerable System Confidentiality Impact; H = High; VI = Vulnerable System Integrity Impact)
There are probably situations where you can have a total technical impact without both of those being High, but I think it's a good first approximation. 

This would also change #195 . 

Probably wait to address this issue until CVSS v4 is finalized. The public comment period on CVSS v4 ends July 31.

j---:
Based on discussion in #284 , I believe this should close #284 if merged.

### DIFF
--- a/docs/topics/information_sources.md
+++ b/docs/topics/information_sources.md
@@ -40,14 +40,29 @@ Three prominent examples are CVSS impact base metrics, CWE, and CPE.
 {% include-markdown "../_includes/_cvss4_question.md" heading-offset=1 %}
 
 [*Technical Impact*](../reference/decision_points/technical_impact.md) is directly related to the CVSS impact metric group.
-However, this metric group cannot be directly mapped to [*Technical Impact*](../reference/decision_points/technical_impact.md) in CVSS version 3  because of the Scope metric.
+The interpretation is different for CVSS version 3 than version 4.
+
+For CVSS v4, the impact metric group can be directly mapped to [*Technical Impact*](../reference/decision_points/technical_impact.md). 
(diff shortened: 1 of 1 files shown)


ahouseholder:
I made a few formatting changes:
- add mapping tables instead of logic statements
- add CVSSv3 subheading to differentiate it from CVSSv4 content
- remove CVSSv4/v3 inset box that is saying "we'll do v4 later", since we're doing this one now.

Screenshots of the current rendering below:
 
Screenshot 2024-03-05 at 1 28 07 PM
Screenshot 2024-03-05 at 1 28 16 PM

j---:
> I made a few formatting changes:
> 

Thanks, these are great.

ahouseholder (approved):
These are ready to go.

PR #533, opened 2024-03-08
Set up site deployment on publish branch

ahouseholder:
This PR changes `.github/workflows/deploy_site.yml` so that it will run on pushes to the `publish` branch. This is needed so that we can deploy the new site to `certcc.github.io/SSVC`


j--- (approved):

sei-vsarvepalli (approved):
lets publish!

PR #549, opened 2024-04-03
Update changelog

ahouseholder:
We didn't update the [changelog](https://certcc.github.io/SSVC/about/changelog/) file to match the [release notes](https://github.com/CERTCC/SSVC/releases/tag/v2024.3) for version 2024.3 last month. This PR fixes that in `main`.

Once this is merged into `main`, I'll do a second PR to merge `main` &rarr; `publish` so the web site will update.

I also plan to move the 2024.3 release tag to point to the updated commit on `main`. Room for debate whether the right thing to do is to (a) move the tag or (b) add a new tag like 2024.3.1 since there have been a few dependency updates. 
I'm ok with either, and neither one needs to be a binding precedent. If you'd prefer option B, please comment as much on this PR, otherwise I'll plan to execute option A.


sei-vsarvepalli (approved):

ahouseholder:
Noting for future reference that I did not, in fact, move the existing 2024.3 tag. Instead, I created a new release `2024.3.1`. It seemed like the more correct thing to do since there were in fact a few small changes beyond just the change log file.

PR #553, opened 2024-04-12
Add link to certcc.github.io home page

ahouseholder:
This PR changes the top-level nav to distinguish the SSVC home page (`index.md`) from the `certcc.github.io` home page.
This will let us provide a way for folks to navigate back and forth across our various `certcc.github.io` sites more easily.


sei-vsarvepalli:
This is a good idea. We should list all the projects eventually under certcc.github.io.

sei-vsarvepalli (approved):

PR #581, opened 2024-06-13
Dockerize unit tests

ahouseholder:
This PR adds a Dockerfile that installs the requirements and runs the included unit tests.

Usage:
```bash
$ docker build -t ssvc_test .
...
$ docker run -it --rm ssvc_test 
............................................................ [100%]
60 passed in 0.71s
```

I'd like to eventually dockerize the dev server too, but that will come later.


sei-vsarvepalli (approved):
Docker works well - useful.

Issue #576, opened 2024-05-30
The JSON for SSVC "options" splits out keys into individual records


jayjacobs:
The schema (with example in `data/schema_examples/Computed-CVE-2014-0751-Coordinator.json`) is being leveraged by CISA for Vulnrichment and creates some unfriendly JSON. 

I opened [issue #40 on vulnrichment](https://github.com/cisagov/vulnrichment/issues/40) to talk about it and they suggested they are just following this schema. 

Long story short, by specifying each key:value pair in it's own object under `options`, it is flattened (by tools) to be unique records, when all of those key:value pairs represent a single object.  (see issue 40 in vulnrichment)

It also does not make sense to have the `options` specified as an array if it is a single object tied to the single `computed` field in the same record. 

Fixed example: 
```
{
    "role": "Coordinator",
    "id": "CVE-2014-0751",
    "version": "2.0.3",
    "computed": "SSVCv2/E:A/V:S/T:T/P:M/B:A/M:M/D:A/2021-09-29T15:29:44Z/",
    "timestamp": "2021-09-29T15:29:44Z",
    "options": {
	"Exploitation": "active",
	"Automatable": "no",
	"Technical Impact": "total",
	"Mission Prevalence": "Minimal",
	"Public Well-being Impact": "Material",
	"Mission & Well-being": "medium"
    },
    "$schema": "https://democert.org/ssvc/SSVC_Computed_v2.0.3.schema.json",
    "decision_tree_url": "https://democert.org/ssvc/CISA-Coordinator-v2.0.3.json"
}
```

sei-vsarvepalli:
Hello Jay,

Thanks for your feedback. I believe the intent in options being an array was for it to be an ordered list, the Decisions Points in schema are considered ordered. This makes it possible to arrange the Decision Point bundles and embed them both in a Computed schema as well the full Decision Tree schema.  There was some discussion related to this earlier that may be relevant 

https://github.com/CERTCC/SSVC/issues/403 
https://github.com/CERTCC/SSVC/discussions/290

Some related information from JSON Schema discussions Google Group

https://groups.google.com/g/json-schema/c/rgkxYocPSVg

jayjacobs:
First, I do not understand the need for these to be an ordered list. The "decision tree" in SSVC is simply for visualizing. The ordering of variables will never change the outcome. But that discussion is not what this issue is about. 

[...]

PR #586, opened 2024-06-25
Make schema available via data/ folder for certcc.github.io

sei-vsarvepalli:
Some small changes in preparation of creation and automation of schema url's and removal of CSS helper view.


ahouseholder (approved):
I'm approving this given that it's necessary to proceed with the effort on #576. 

However, we _might_ want to consider whether softlinking all of `/data` from `/docs/data` is the right approach. I think it's probably okay, but we might come across a reason to be more restrictive than that in the future. (E.g., maybe only linking `/data/schema/` to `/docs/data/schema/`) But I don't have a strong enough argument against it to change that right now.

PR #587, opened 2024-06-25
Merge main --> publish

ahouseholder:
Catch publish up with main so that #586 is live


sei-vsarvepalli (approved):
Some files are superfluous here, like Dockerfile, but perhaps irrelevant. Just capturing for my notes really in the future.

Issue #454, opened 2024-02-09
Consider versioning Outcomes like Decision Points


ahouseholder:
Expanding on:

- #289
- #388 
- #445 

It seems like we should consider versioning Outcome sets as well. This seems like a relatively easy thing to clone from what we did in #350 and #368 for decision points.

j---:
Do outcome sets follow the same versioning rules as decision points, or as decision point groups? I guess outcomes are analogous to decision points and outcome sets are analogous to decision point groups?

ahouseholder:
See also #745

ahouseholder:
Resolved by #791

PR #588, opened 2024-06-25
Feature/issue 576

### LINKED ISSUE #576: The JSON for SSVC "options" splits out keys into individual records

sei-vsarvepalli:
This is a major change in Schemas to resolve #576  and try to create new schemas that can be processed by third parties efficiently. This requires some changes to [vulnrichment](https://github.com/cisagov/vulnrichment)  project with ADP exports. Our schema has been updated, but there are external dependencies.

https://github.com/cisagov/vulnrichment/issues/40

There is a new option in the schema called `outcome` which is optional and can be used represent the comprehensive outcomes of the current Decision Tree. This can provide outcomes of the full decision tree, just like the outcomes from a decision point, so the next consumer can use this values to restrict outputs from the decision tree.

The JavaScript code also has been updated. This requires a bit of thorough review as there may be other dependencies that are impacted by this schema comprehesnive update.  Currently there is a version number in the Computed Schema, but that should be managed before parsers are written to read Computed schema when used alongside of ADP or other data providers.

@j---  may need to help coordinate with CISA developers to ensure the new schema is adopted by CISA developers.


data/schema_examples/Computed-CVE-2014-0751-Coordinator.json
```diff
@@ -1,29 +1,17 @@
...
+    "options": 	{
+	    "Exploitation": ["active"],
+	    "Automatable": ["no"],
+	    "Technical Impact": ["total"],
+	    "Mission Prevalence": ["Minimal"],
+	    "Public Well-being Impact": ["Material"],
+	    "Mission & Well-being": ["medium"]
+    },
+    "outcomes": ["Act"],
     "decision_tree_url": "https://democert.org/ssvc/CISA-Coordinator-v2.0.3.json"
```
ahouseholder, line 16: See previous comment about removing CISA tree

data/schema/SSVC_Computed.schema.json
ahouseholder, line 1: General comment on the Computed Schema: is there a way for us to have the computed schema directly draw on the `Decision_Point.schema.json` or possibly the `Decision_Point_Group.schema.json`? I'll elaborate my thinking in a comment on the PR.

ahouseholder:
Following up on [this comment](https://github.com/CERTCC/SSVC/pull/588#discussion_r1655428766) and expanding on [this comment](https://github.com/CERTCC/SSVC/issues/576#issuecomment-2189231325) from #576:

It seems like we need the list of chosen items to include more specificity about what decision point they're drawn from. I was thinking something like:

```json
{
  "decision_points": [
    {
      "namespace": "ssvc",
      "name": "Exploitation",
      "version": "1.0.0",
      "key": "E",
      "values": [ "None" ]
    },
    {
      "namespace": "ssvc",
      "name": "Automatable",
      "version": "2.0.0",
      "key": "A",
      "values": [ "Yes" ]
    }
  ]
}
```

Which shares the `namespace`, `name`, `version`, `key`, and `values` properties from the DP schema but only includes the `name`(s) of the chosen `values`. (Note that `description` is present in the DP schema but omitted in the example, as I don't think it's necessary in this case, we're not defining the decision point with this record we're just downselecting from the available options. Since each decision point has a namespace, a name, and a version, it seems important to fully identify them though.

In the current Computed example we have

```json
{
"Exploitation": "active"
},
{
  "Automatable": "no"
}
```

which is now underspecified since there can be different versions of _Exploitation_ and _Automatable_.

data/schema/SSVC_Computed.schema.json
sei-vsarvepalli: Yeh - Computed Schema should tie to a Decision Point Group Schema, so the plan was to hav ea pedigree like version tracking

Decision Points (multiple)  => Decision Point Group (Set selected and a combined versioned) => Provisioned Schema (DP's ordered and versioned) => Computed Schema (tied to Provision version )

This may need a short discussion, I can see it is not clear in the Schema docs, it should be more explicit that DP_Group is tied to a Provisioned schema with a specific version.

sei-vsarvepalli:
> Following up on [this comment](https://github.com/CERTCC/SSVC/pull/588#discussion_r1655428766) and expanding on [this comment](https://github.com/CERTCC/SSVC/issues/576#issuecomment-2189231325) from #576:
> 
> It seems like we need the list of chosen items to include more specificity about what decision point they're drawn from. I was thinking something like:
> 
> ```json
> {
>   "decision_points": [
>     {
>       "namespace": "ssvc",
>       "name": "Exploitation",
>       "version": "1.0.0",
>       "key": "E",
>       "values": [ "None" ]
>     },
>     {
>       "namespace": "ssvc",
>       "name": "Automatable",
>       "version": "2.0.0",
>       "key": "A",
>       "values": [ "Yes" ]
>     }
>   ]
> }
> ```
> 
> Which shares the `namespace`, `name`, `version`, `key`, and `values` properties from the DP schema but only includes the `name`(s) of the chosen `values`. Note that `description` is present in the DP schema but omitted in the example, as I don't think it's necessary in this case, we're not defining the decision point with this record we're just downselecting from the available options. Since each decision point has a namespace, a name, and a version, it seems important to fully identify them though.
> 
> In the current Computed example we have
> 
> ```json
> {
> "Exploitation": "active"
> },
> {
>   "Automatable": "no"
> }
> ```
> 
> which is now underspecified since there can be different versions of _Exploitation_ and _Automatable_.

Optionally the full Decision Tree ordered with outcomes (compliant to `Provisioned` schema) can be embedded to show all the information. In the compressed form, the chained version pedigree will be the way to be clear about what DP's are being used.

sei-vsarvepalli:
Some more issues to be concerned about 

- Depends on  #454 
- Outcome groups should be a schema perhaps ? #589 
- Small bug src/ssvc/policy_generator.py PolicyGenerator sample fix AUTOMATABLE_1 to be AUTOMATABLE_2 #590 
- Should Policy have a schema that consumes Outcome Groups and should it be versioned? #591 #592 
- Provision schema is essentially a full Policy, which is a full Decision Tree. If Policy is versioned, there is no concern about versioning this distinctly, if not this may also need versioning 
- Computed schema is essentially to communicate an evaluated instance of a Policy for a specific vulnerability or group of vulnerabilities. It should also be versioned and can optionally include the full Policy.
- All schemas should have a version as well to ensure any new JSON schema - may also need distinct $id fields and folder/locations in url.

sei-vsarvepalli:
The example 
https://github.com/CERTCC/SSVC/blob/7d5aadf71fec2f388f483d50856ab5384a665c8f/data/schema_examples/Decision_Point_Group_Selection.json
 does not have an identifier such as a CVE or VU# or GHSA etc. This needs to be discussed further too..

ahouseholder:
I think we've already addressed the base issue #576 sufficiently, and the other issues in this PR were split up (see comments above). So I'm closing this PR as unmerged. Smaller chunks of it might still be relevant to future PRs of course.

PR #597, opened 2024-07-09
Update mkdocs.yml

ahouseholder:
Remove link to polyfill.io 
See https://github.com/squidfunk/mkdocs-material/issues/7295
and https://censys.com/july-2-polyfill-io-supply-chain-attack-digging-into-the-web-of-compromised-domains/


sei-vsarvepalli (approved):
yes.

PR #599, opened 2024-07-09
Add Decision Point Value Selection schema with an example

sei-vsarvepalli:
This will be an attempt to breakup our work from schema development that will help #588  into smaller pieces and run test cases against this.

All the future schemas will start livings in `data/schema/current/` folder symlinked to a schemaVer version controlled schema documents.


data/schema/current/Decision_Point_Group_Selection.schema.json
ahouseholder, line 1: Suggest renaming to `Decision_Point_Value_Selection.schema.json`

data/schema/v1/Decision_Point_Group_Selection-1-0-1.schema.json
ahouseholder, line 1: Suggest rename to `Decision_Point_Group_Selection-1-0-1.schema.json`

data/schema_examples/CVE-1969-0000-Decision_Point_Group_Selection.json
ahouseholder, line 1: Suggest rename to `CVE-1969-0000-Decision_Point_Value_Selection.json`

docs/ssvc-calc/Decision_Point.schema.json
ahouseholder, line 1: Is it necessary to have a separate set of schemas in `docs/ssvc-calc`? Or could they be included in `data/schema`? (If these are the same as the ones already in `data/schema`, can they be replaced by softlinks? Just trying to avoid having to maintain two copies of the same file. 

Might be a good time to add a version string to them while we're touching them too.

docs/ssvc-calc/Decision_Point_Group.schema.json
ahouseholder, line 1: See comment on previous file: 

- Can this be replaced by a link?
- Should it get a version string while we're at it?

ahouseholder (changes requested):
I think the filenames are overloading the phrase "Decision Point Group" which has meaning in the definition of decision models. Can we call it "Decision Point Value Selection" instead? 

I'd like us to keep this schema as absolutely narrowly scoped as possible. This schema does not require the creation of a "decision point group" object, as it's just a list of decision points and selected values from them.

docs/ssvc-calc/Decision_Point_Group.schema.json
sei-vsarvepalli: These two files are spurious - I can remove these really.. Definitely not focus of this PR

data/schema_examples/CVE-1969-0000-Decision_Point_Group_Selection.json
```diff
@@ -0,0 +1,32 @@
+{
+    "id": "CVE-1969-0000",
+    "role": "Coordinator",
```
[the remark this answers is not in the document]
sei-vsarvepalli: Do you mean to remove it from the "example" or also from the schema? 

As I see it: 
Role is just an optional field anyway. I think it is useful if CISA wants to publish and there is another stakeholder type also wants to publish SSVC metrics - we can distinguish the type of stakeholder.

docs/ssvc-calc/Decision_Point.schema.json
sei-vsarvepalli: These two files are spurious - I can remove these really.. Definitely not focus of this PR

data/schema_examples/CVE-1969-0000-Decision_Point_Group_Selection.json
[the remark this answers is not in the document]
sei-vsarvepalli: BTW Do we also need SSVC Vector Representation as part of the Group Selection schema?

sei-vsarvepalli:
Looks like the data/schema/Decision_Point.schema.json and data/schema/Decision_Point_Group.schema.json should be setup for some Test Cases to run properly. adding these with some vision strings and tests to run properly

sei-vsarvepalli:
I think we are mostly there for Value Selection schema which is ideally what ADP like capabilities can use to reliably provide their evaluation of a vulnerability using SSVC. Now some questions remains
1. If SSVC is versioned and all the Decision Points are fixed that are part of SSVC 2.1 and are documented. Is it redundant or over-stating to provide version under each Decision Point ?
2. Should Value Selection schema have a Vector representation that simplifies a specific evaluation? 

It may be helpful for @j---  to also comment about these questions for operational viability and usage in CISA. Issue #576 will get a very different answer if this PR is merged.

ahouseholder:
> 1. If SSVC is versioned and all the Decision Points are fixed that are part of SSVC 2.1 and are documented. Is it redundant or over-stating to provide version under each Decision Point ?

The version in each decision point is the decision point version, which we describe in https://certcc.github.io/SSVC/adr/0006-ssvc-decision-point-versioning-rules/

That's not the same as SSVC 2.1 (which FWIW we're already past with v2024.3) -- but the idea is that eventually a decision point version approaches stability even as the rest of SSVC continues to be revised. We really don't want to get into having to descramble SSVC v2.1's Exploitation vs SSVC v2024.3's Exploitation vs SSVC v2024.3.1's Exploitation vs SSVC v2024.3.2's Exploitation. Because across all extant SSVC versions, Exploitation has only changed twice ([Exploitation v1.0.0 and v1.1.0](https://github.com/CERTCC/SSVC/blob/21a0bdcda2bbc3b8dd3bb64d00e76e0ba70f467f/src/ssvc/decision_points/exploitation.py#L53-L75))

ahouseholder:
> 2. Should Value Selection schema have a Vector representation that simplifies a specific evaluation?

I'm of the opinion that the answer is likely "No" based on

- #333
- #451 

where we took the vector stuff out of everything except the calculator.  
If we need to reopen the decision about the vector, I'd suggest we treat it as a separate issue/PR from this one.

sei-vsarvepalli:
Another question came up in my mind - should there a optional "references" section in Selection schema to provide information such as Public POC was seen at a URL with optional "type" of reference information such as "exploit" - say something like

```
{"references" : [
      { 
          "url": "https://www.exploit-db.com/exploits/32998",
          "type": "exploit"
        } 
]}

```

data/schema/v1/Decision_Point-1-0-1.schema.json
ahouseholder, line 1: I'm almost certainly overthinking this, but does using SemVer for schema versioning mean we need to define versioning rules? Might CalVer (`YYYY-MM` with optionally `YYYY-MM-patch`) be preferable so we can avoid having to say much more than "subsequent versions will be dated"?

ahouseholder (approved):
I don't have any objections to proceeding with this PR as-is. My comment regarding versioning rules for schemas isn't a reason to hold this up. I'm planning to spawn it into an issue so we can at least track the outcome of any ensuing discussion.

ahouseholder:
> Another question came up in my mind - should there a optional "references" section in Selection schema to provide information such as Public POC was seen at a URL with optional "type" of reference information such as "exploit" - say something like
> 
> ```
> {"references" : [
>       { 
>           "url": "https://www.exploit-db.com/exploits/32998",
>           "type": "exploit"
>         } 
> ]}
> ```

I'd like to split this off into a separate issue

- #600 

because it seems like we should consider whether the references apply to the list or to the individual decision point selections. I could imagine you'd want to say "we chose `foo` for Decision Point `A` because `reference` and `bar` for Decision Point `B` because `other reference`. We might need to understand better if/how this would be used to make a decision.

data/schema/v1/Decision_Point-1-0-1.schema.json
sei-vsarvepalli: Yeh - I think we can move that way later without a problem. The idea would be to provide "breaking" vs "non-breaking" Schema changes to be version controlled.  Mostly from [Snowplow's musings](https://docs.snowplow.io/docs/pipeline-components-and-applications/iglu/common-architecture/schemaver/) . 

The challenge with some of these issues seems to show that much of the schema versioning itself is so experimental and we are trying to imagine an operational viability which is sort of impossible to know till we have enough usage.

ahouseholder: Ok, good enough for now. We should generate an ADR pointing to the snowplow link. I'll spawn an issue for that.

data/schema_examples/CVE-1969-0000-Decision_Point_Group_Selection.json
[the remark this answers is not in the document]
ahouseholder: Role being optional per the schema seems reasonable.

My hunch is that folks will look at the example more than the schema and so if the example included it that it would turn into an expectation even when it's not required. And since I can think of situations where it might not make sense, I was concerned about introducing potential confusion in cases where you just want to share some facts about a decision point you looked at.

See [this comment](https://github.com/CERTCC/SSVC/pull/599#issuecomment-2223477298) for thoughts on the vector question.

Issue #602, opened 2024-07-11
Add schemaVersion field as required to Decision Point schema and Decision Point Group schema for the next version. 


sei-vsarvepalli:
Mostly this is just related to PR #599 as a follow-up the Decision Point and Decision Point Group schema both needs to have the schemaVersion field added as well as code in test_schema.py updated to ensure the schemaVersion is being used in all the various output Schema JSON files.

Vijay

PR #598, opened 2024-07-09
Update dependencies

ahouseholder:
This PR is to publish the Current main branch

PR #603, opened 2024-07-11
Update dependencies (#598)

sei-vsarvepalli:
Push new schemas to production URL's - in preparation for #602 work.


sei-vsarvepalli:
Actually a small error in the old PR, we need to added schemaVersion as a possible field first. closing this one for now.

PR #604, opened 2024-07-11
Feature/issue 602

sei-vsarvepalli:
This is the first stage of two stage PR's to resolve adding schemaVersion field into the Decision Point and Decision Point Group schemas. Whole the schemaVersion was in the "definitions" it was not part of the optional field. This needs to be fixed first and pushed to "main" and the "public" branches so the $id based URL's are accessible for the test cases to run properly.

Vijay


sei-vsarvepalli:
Strange this PR is broken.

PR #606, opened 2024-07-11
Resolve part of Feature/issue 602 to add schemaVersion field

### LINKED ISSUE #602: Add schemaVersion field as required to Decision Point schema and Decision Point Group schema for the next version.

sei-vsarvepalli:
We also need to update the schema files to allow for schemaVersion to be an available field. This may require later changes to all the data/json/decision_points/*.json files also. But moving these changes slowly to ensure we have something in "public" branch that exposes remote URL's for schema validations.

- resolves #602

### DIFF
--- a/data/json/decision_points/automatable_2_0_0.json
+++ b/data/json/decision_points/automatable_2_0_0.json
@@ -1,7 +1,7 @@
 {
-  "schemaVersion": "1-0-1",
   "namespace": "ssvc",
   "version": "2.0.0",
+  "schemaVersion": "1-0-1",
   "key": "A",
   "name": "Automatable",
   "description": "Can an attacker reliably automate creating exploitation events for this vulnerability?",
--- a/data/json/decision_points/automatable_2_0_0.json
+++ b/data/json/decision_points/automatable_2_0_0.json
@@ -1,4 +1,5 @@
 {
+  "schemaVersion": "1-0-1",
   "namespace": "ssvc",
   "version": "2.0.0",
   "key": "A",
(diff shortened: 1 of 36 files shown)


sei-vsarvepalli:
the additional generated JSON looks fine. One assumption that I have made is when schemaVersion is updated to 1-0-2, all the schemas (3 at this time but may be as many as 6 possible schemas) may need to get a new release. This means it is possible that we only changed Decision_Point_Value_Selection schema to add `references` field, but all other schemas will also say get an 1-0-2 schemaVersion update. This is how the code is written, will this be a concern? We can override it per schema, there are benefits for one way or another of doing this schemaVersion which we should discuss.

ahouseholder:
> One assumption that I have made is when schemaVersion is updated to 1-0-2, all the schemas (3 at this time but may be as many as 6 possible schemas) may need to get a new release. This means it is possible that we only changed Decision_Point_Value_Selection schema to add `references` field, but all other schemas will also say get an 1-0-2 schemaVersion update. This is how the code is written, will this be a concern? We can override it per schema, there are benefits for one way or another of doing this schemaVersion which we should discuss.

I think this is okay for now (as in, I'm okay with waiting until it's actually a problem for us to make it a problem).

ahouseholder (approved):

Issue #640, opened 2024-10-28
CSAF / SSVC integration


ahouseholder:
**Describe the solution you'd like**

This issue is for tracking integration with CSAF and SSVC

See also 

- OASIS CSAF issue: oasis-tcs/csaf#803
- https://groups.oasis-open.org/discussion/motion-for-803#bm87c96b9f-de9e-4b75-9a60-25f534cfbbda

tschmidtb51:
@ahouseholder: Sorry for the late reply.
Unfortunately, #642 did not completely change the schema base. Here is the issues that I see so far:
- [ ] `definitions` is not longer valid. It needs to be changed to `$defs`

PR #642, opened 2024-10-30
Change json schema base to 2020-12

sei-vsarvepalli:
This PR modifies the base url for the JSON schema to change from `draft-07` to `2020-12`

- https://json-schema.org/draft/2020-12/release-notes

Original text:

> This will be a good one to merge to Main and then finally publish with all the latest changes. I don't think any version  number changes are needed as the schema is entirely the same and works fine with both draft-07 and 2020-12 the same way.
> Vijay


ahouseholder:
This PR is in support of

- #640

ahouseholder (approved):

Issue #644, opened 2024-10-31
Looks like tab-based navigation has changed breaking some of the content navigation between JSON and text.


sei-vsarvepalli:
**Describe the bug**
Visit https://certcc.github.io/SSVC/reference/decision_points/supplier_engagement/ and see the tab-based navigation for `Text` and `JSON` fields. 

**To Reproduce**
Click on `Text` and `JSON` tabs under Supplier Engagement in this example

**Expected behavior**
Expect tab-based navigation to work. 

**Screenshots**
![Screenshot 2024-10-31 at 1 24 28 PM](https://github.com/user-attachments/assets/043026a7-626b-4b24-90a9-89c487cd7cfd)
vs 
![Screenshot 2024-10-31 at 1 24 19 PM](https://github.com/user-attachments/assets/ea5ccc6b-dca4-411f-9044-65813e39c65e)

**Platform details**

**Additional context**
I think the medics squidfunk libraries have changed quite a bit - this may have been broken for a few months from previous TypeScript library for squidfunk changes.

ahouseholder:
Might be something that changed with 

- #638

Those pages use the include-markdown plugin

ahouseholder:
> Indented code blocks must be surrounded by newlines
Now mkdocs-include-markdown-plugin will only detect indented code blocks if are surrounded by newlines, conforming to [CommonMark specification](https://spec.commonmark.org/0.28/#indented-code-blocks).

> In the practice this means that you must surround indented code blocks with newlines or possible link targets URLs will be rewritten to work in relative files.

ahouseholder:
Resolved by #645

PR #645, opened 2024-10-31
Updates from mkdocs latest version for space issues

sei-vsarvepalli:
This is to fix the website issue, it looks like bug was introduce after October 1 from the commit hash c54c444897481af0fcf2c12f8d92765896e149d3 which is the last working version of medics to website for tab-based navigation for this issue #644 

This primarily targets the `publish` branch eventually for fixing this issue that makes the website look wonky.


ahouseholder (approved):
Tested this locally after updating dependencies to current. It looks good to me.

PR #654, opened 2024-11-27
SSVC JSON Schema 2020-12

tschmidtb51:
- change `definitions`to `$defs`
- fixes https://github.com/CERTCC/SSVC/issues/640#issuecomment-2504366097

### DIFF
--- a/data/schema/v1/Decision_Point-1-0-1.schema.json
+++ b/data/schema/v1/Decision_Point-1-0-1.schema.json
@@ -3,7 +3,7 @@
     "title": "Decision Point schema definition",
     "$id": "https://certcc.github.io/SSVC/data/schema/v1/Decision_Point-1-0-1.schema.json",
     "description": "Decision points are the basic building blocks of SSVC decision functions. Individual decision points describe a single aspect of the input to a decision function.",
-    "definitions":     {
+    "$defs":     {
 	"schemaVersion": {
             "description": "Schema version used to represent this Decision Point",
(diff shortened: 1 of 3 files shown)


sei-vsarvepalli:
Hello @tschmidtb51 


I believe we have avoided $defs due to some limitations with CVE program not being able to have $ character in their schemas due to some restrictions in AWS implementation of MongoDB.

https://docs.aws.amazon.com/documentdb/latest/developerguide/functional-differences.html#functional-differences.field-name-restrictions

https://github.com/CVEProject/cve-schema/issues/144

Do you require $defs specifically or is it a design preference?  Let us know we are happy to support what will work for multiple consumers that can adopt our metrics.

Thanks
vijay

sei-vsarvepalli:
It looks like we are using $refs anyway as a key. May need to discuss with CVE schema to make sure there are no issues/concerns. And it looks like the 

https://json-schema.org/draft/2020-12/schema

states 

```
        "definitions": {
            "$comment": "\"definitions\" has been replaced by \"$defs\".",
            "type": "object",
            "additionalProperties": { "$dynamicRef": "#meta" },
            "deprecated": true,
            "default": {}
        }
        
```

sei-vsarvepalli:
Yeh after looking through Amazon documentation and CVE AWG implementation, this is not a concern. It looks like it is only a problem for generated and stored values from the schema that cannot have $ sign in their key values. So I will go ahead and approve the change.

PR #669, opened 2025-01-22
fix unit test

ahouseholder:
The JSON schema unit tests were failing because the local path substitution wasn't working consistently across different environments. This PR fixes that by computing a relative path from the test file in which it is called.

### DIFF
--- a/src/test/test_schema.py
+++ b/src/test/test_schema.py
@@ -13,33 +13,38 @@
 
 import json
 import logging
+import os
 import unittest
 
 import jsonschema
 from jsonschema import Draft202012Validator
 from referencing import Registry, Resource
-import os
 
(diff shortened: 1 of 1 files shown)


sei-vsarvepalli (approved):
See #671  comment for the test scripts fix.

ahouseholder:
Because our schemas are nested (they have internal pointers to other schemas we also run out of this repo), we can't just say "load the schema from a local file path" and be done with it. We appear to need the retrieval intercept that rewrites our URIs to file paths and loads from there. There might be something more we could do with the schema validator object, but that's going deeper into the validator API than I think we need to at this point. 

I'm going to merge this since the tests work as-is. If we need to improve the tests, let's do a separate issue for that.

PR #671, opened 2025-01-22
Reformat python with black

ahouseholder:
Doing this as a bulk change to catch up older code with current formatting. Just concentrating a bunch of format changes into a single PR.


src/test/test_schema.py
```diff
@@ -27,11 +27,17 @@
...
+    CVSSv3,
+    CVSSv4,
+)  # noqa
 
 # importing these causes the decision points to register themselves
 from ssvc.dp_groups.ssvc.collections import SSVCv1, SSVCv2, SSVCv2_1  # noqa
 
+
 def retrieve_local(uri):
     fileuri = uri.replace("https://certcc.github.io/SSVC", os.getcwd())
```
sei-vsarvepalli, line 42: If we are fixing the tests, we can make this an environment variable so local mkdocs/docker can do this test without the remote dependency?

sei-vsarvepalli: I mean make the var `https://certcc.github.io/SSVC` from an environment variable, that is set when running tests.

ahouseholder: I'll address this thread in #669

PR #672, opened 2025-01-23
Add versioned outcomes schema

### LINKED ISSUE #589: Need json schema for outcome groups

We have json schemas for decision points and decision point groups but not for outcome sets. We need that.

- offshoot of #576

sei-vsarvepalli:
This is an attempt to move towards Policy Explorer which requires Outcomes and resolves #589  so we have a schema for the outcome groups.

Here is what has happened 
1. Outcomes Schema description under data/schema/v1 and data/schema/current (symlink)
2. Outcomes JSON files under data/json/outcomes
3. Create CVSS Vectors that were predefined to create JSON format in data/json/decision_points/cvss/ 
4. Scripts to create CVSS decision_points and Outcomes in data/json/outcomes

Mostly additions should not disturb any setup.


data/json/decision_points.zip
ahouseholder, line 1: Do we need this zip file for some reason? It seems like it's redundant to `data/json/decision_points/*`

src/cvss_to_json.py
```diff
@@ -0,0 +1,22 @@
...
+        "target_distribution", "user_interaction"]
+for mod in mods:
+    module = getattr(__import__('ssvc.decision_points.cvss', fromlist=[mod]),
+                     mod)
+    for dp in dir(module):
+        if dp.upper().find(mod.upper()) > -1:
+        #user_interaction USER_INTERACTION_2
+            print(mod, dp)
+            sdp = getattr(module, dp)
+            f = open(f"../data/json/decision_points/cvss/{dp.lower()}.json", "w")
```
ahouseholder, line 19: Purely a syntactical note, not worth changing, just mentioning it since I noticed it here.

```
with open("...","w") as f:
    f.write("foo")
```

does the same thing but you don't have to explicitly `close()` because the context manager handles it for you.

src/cvss_to_json.py
ahouseholder, line 1: Just a comment on future plans: I eventually want to make this generation stuff happen as part of the site build process automatically. This will come in handy for that so we don't have to remember to refresh it manually.

src/ssvc/gini_impurity.py
ahouseholder, line 1: Why is this file here? It seems like it's from a tutorial about sklearn?

ahouseholder (changes requested):
I'm fine with the bulk of this PR as-is. My only to blockers to merge are:

- [ ] do we need `data/json/decision_points.zip`?
- [x] do we need `src/ssvc/gini_impurity.py`

both of these seem out of place to me.

ahouseholder:
Also note that we'll need to do a brief reconciliation of #674 with this one, just depends on merge order which one will need to change.

src/cvss_to_json.py
sei-vsarvepalli: yeh - that was the idea to keep my scripts there for later absorption or attaching to the .github/workflows

data/json/decision_points.zip
sei-vsarvepalli: nop e can be removed. I will update the PR.

src/ssvc/gini_impurity.py
sei-vsarvepalli: yeh - sorry can remove those two files..

src/cvss_to_json.py
sei-vsarvepalli: will do with syntax...

ahouseholder (approved):

PR #674, opened 2025-01-24
Replace python dataclasses with pydantic dataclasses

### LINKED ISSUE #673: Switch from python dataclasses to pydantic BaseModel dataclasses

We will soon be working on some FastAPI-based microservices using the `ssvc` library, and it's a little easier for us if we can be consistent with what kind of dataclasses we're supporting in the modules. Having used pydantic dataclasses inheriting from `pydantic.BaseModel` in other projects, (and the fact that FastAPI supports pydantic classes natively), this seems like a good time to make the shift.

**Describe alternatives you've considered**

Staying with the existing python dataclasses was the other option, but pydantic is superior because:

- pydantic's validators are better
- json round-trip is built-in rather than being an add-on
    - allows us to remove a dependency on `dataclasses-json`
- FastAPI already uses pydantic dataclasses, so we don't need to maintain two different dataclass implementations in the same project

ahouseholder:
- resolves #673 

Rationale:

- sets us up for future FastAPI microservice development
- removes dependency on `dataclasses-json`
- improves our ability to use validators on object creation


src/cvss_to_json.py
```diff
@@ -1,22 +1,57 @@
...
+        module = getattr(__import__("ssvc.decision_points.cvss", fromlist=[mod]), mod)
+        for dp in dir(module):
+            if dp.upper().find(mod.upper()) > -1:
+                # user_interaction USER_INTERACTION_2
+                print(mod, dp)
+                sdp = getattr(module, dp)
+                with open(
+                    f"../data/json/decision_points/cvss/{dp.lower()}.json", "w"
+                ) as f:
+                    f.write(sdp.model_dump_json(indent=2))
```
ahouseholder, line 53: I realized I also had to fix these because `to_json` is now `model_dump_json`, and wrapped them in a `main()` func while I was here

src/outcomes_to_json.py
```diff
@@ -1,11 +1,29 @@
...
-  with open(f"../data/json/outcomes/{x}.json","w") as f:
-   f.write(outcome.to_json())
 
- 
+def main():
+    for x in dir(groups):
+        outcome = getattr(groups, x)
+        if type(outcome) == OutcomeGroup:
+            with open(f"../data/json/outcomes/{x}.json", "w") as f:
+                f.write(outcome.model_dump_json(indent=2))
```
ahouseholder, line 25: I realized I also had to fix these because `to_json` is now `model_dump_json`, and wrapped them in a `main()` func while I was here

ahouseholder (commented):
This whole commit was just me regenerating the json files with the fixed scripts

sei-vsarvepalli (approved):
Pydantic it is now.

PR #683, opened 2025-02-17
Convert site to use inline python instead of relying on manual script to generate decision point examples

### LINKED ISSUE #431: Replace doctools.py with embedded python examples
We currently have a separate tool `doctools.py` that generates a bunch of `.md` and `.json` files for the various decision points, outcome sets, etc. This works ok, assuming we remember to run it after changing decision points, outcome sets, etc. As you might imagine, this is not an obvious thing to do, so it's likely that the docs will fall out of sync with the code.

The `markdown-exec` plugin provides the ability to run python from within an example block. https://pypi.org/project/markdown-exec/

My proposal is that we convert the current generated `.md` and `.json` files that are subsequently included in other markdown to be just normal markdown files that directly call the python to generate the markdown table and .json examples at site build time.

I have similar code already working in the Vultron project. For example, all the `.json` blobs in https://certcc.github.io/Vultron/howto/activitypub/activities/manage_case/ are created this way.

It's a cleaner solution and it eliminates the need to remember to do an extra step when certain files change. As a bonus, it is also an opportunity to add unit tests to test the generators too.

ahouseholder:
This PR:

- resolves #431 
- eliminates site dependency on `doctools.py` in favor of direct inline python in the markdown that runs at site deploy time. This change ensures that examples in the site are always the latest available version of any given decision point.
- changes `doctools.py` as follows:
   - only generates the `data/json/decision_points/**/*.json` files 
   - do not generate `docs/_generated/**/*.md` since these are replaced by the aforementioned inline python
- Applies a consistent usage of `VERSIONS` to contain an ordered tuple of all decision point versions defined in a file
- Applies a consistent usage of `LATEST` to be the most recent version of the decision point (last item in the `VERSIONS` tuple)


sei-vsarvepalli (changes requested):
Not a big deal but warning in pedantic seems to indicate that Decision Point Values are expected to be "list" and not "tuple"

https://github.com/CERTCC/SSVC/blob/1bb505002db0bb59f23331a2a97be8fdb96a4d56/src/ssvc/decision_points/cvss/helpers.py#L37

https://github.com/CERTCC/SSVC/blob/1bb505002db0bb59f23331a2a97be8fdb96a4d56/src/ssvc/decision_points/cvss/helpers.py#L101

Changing the above two lines in CVSS from "tuple" to "list" - causes the warnings to go away. I have not thoroughly reviewed the scripts but just going after the warnings I saw in "Checks" tab


that say
```

src/test/test_schema.py::MyTestCase::test_decision_point_group_validation
  /opt/hostedtoolcache/Python/3.10.16/x64/lib/python3.10/site-packages/pydantic/main.py:477: UserWarning: Pydantic serializer warnings:
    Expected `list[SsvcDecisionPointValue]` but got `tuple` with value `(SsvcDecisionPointValue(k...entation for details.'))` - serialized value may not be as expected
    ```

ahouseholder:
> Not a big deal but warning in pedantic seems to indicate that Decision Point Values are expected to be "list" and not "tuple"
> 
> https://github.com/CERTCC/SSVC/blob/1bb505002db0bb59f23331a2a97be8fdb96a4d56/src/ssvc/decision_points/cvss/helpers.py#L37
> 
> https://github.com/CERTCC/SSVC/blob/1bb505002db0bb59f23331a2a97be8fdb96a4d56/src/ssvc/decision_points/cvss/helpers.py#L101
> 
> Changing the above two lines in CVSS from "tuple" to "list" - causes the warnings to go away. I have not thoroughly reviewed the scripts but just going after the warnings I saw in "Checks" tab
> 
> that say
> 
> ```
> 
> src/test/test_schema.py::MyTestCase::test_decision_point_group_validation
>   /opt/hostedtoolcache/Python/3.10.16/x64/lib/python3.10/site-packages/pydantic/main.py:477: UserWarning: Pydantic serializer warnings:
>     Expected `list[SsvcDecisionPointValue]` but got `tuple` with value `(SsvcDecisionPointValue(k...entation for details.'))` - serialized value may not be as expected
>     ```
> ```

Fixed in https://github.com/CERTCC/SSVC/pull/683/commits/f7ffc29e7b18ab491e1707b5d44c843e6d04a880

sei-vsarvepalli (approved):
All looks good.

PR #704, opened 2025-02-20
Update schema JSON to use cross-referencing and force more validations

### LINKED ISSUE #697: SSVC should not be allowed to be empty
**Describe the bug**
Currently, the following is allowed by the DecisionPointSelectionSchema:

```json
{
    "id": "",
    "role": "",
    "selections": [
       {
           "name": "",
            "namespace": "",
            "values": [
                    ""
              ],
               "version": ""
           }
        ],
        "schemaVersion": "1-0-1",
         "timestamp": ""
} 
```

**Expected behavior**
Have at least one character in each string. This can be achieved by adding `"minLength": 1,` to all `string` objects. 

**Additional context**
In CSAF, we just use the rule "If it's there, it has content - otherwise skip the field".

sei-vsarvepalli:
This is starting attempt to resolve #697 - don't merge yet. I am running a number of tests locally as well with this schema change. It is so far a non-breaking change of the schema tools.

- resolves #697

## Copilot summary

This pull request includes several updates to JSON schema files to improve descriptions, add new constraints, and enhance the clarity of the schema definitions. The changes also include updates to test files to reflect the schema modifications.

### Schema improvements:

* [`data/schema/v1/Decision_Point-1-0-1.schema.json`](diffhunk://#diff-37fef1e05e020eda394600d8c7c4b8c22add53a9275d5ca0106313502c8bf9c3L8-R8): Enhanced descriptions for properties, added `minLength` constraints, and included example values for better clarity. Updated the `namespace` property to include a pattern and examples. [[1]](diffhunk://#diff-37fef1e05e020eda394600d8c7c4b8c22add53a9275d5ca0106313502c8bf9c3L8-R8) [[2]](diffhunk://#diff-37fef1e05e020eda394600d8c7c4b8c22add53a9275d5ca0106313502c8bf9c3L18-R32) [[3]](diffhunk://#diff-37fef1e05e020eda394600d8c7c4b8c22add53a9275d5ca0106313502c8bf9c3L44-R79)
* [`data/schema/v1/Decision_Point_Group-1-0-1.schema.json`](diffhunk://#diff-5e95e7e21565f11b6285f23245dbaef8ca1080b4c4b9808289652f2005392209L3-R8): Added a description for the schema, improved property descriptions, and added `minLength` constraints with examples. [[1]](diffhunk://#diff-5e95e7e21565f11b6285f23245dbaef8ca1080b4c4b9808289652f2005392209L3-R8) [[2]](diffhunk://#diff-5e95e7e21565f11b6285f23245dbaef8ca1080b4c4b9808289652f2005392209L19-R32)
* [`data/schema/v1/Decision_Point_Value_Selection-1-0-1.schema.json`](diffhunk://#diff-920f3f6d2bf1d3d7a9c07583e2e8bdc0225cd5149a031641a99993d1b9f442d1R4-R42): Added a schema description, refined property descriptions, added `minLength` constraints, and used `$ref` to reference properties from other schemas. [[1]](diffhunk://#diff-920f3f6d2bf1d3d7a9c07583e2e8bdc0225cd5149a031641a99993d1b9f442d1R4-R42) [[2]](diffhunk://#diff-920f3f6d2bf1d3d7a9c07583e2e8bdc0225cd5149a031641a99993d1b9f442d1L75-R69)

### Test updates:

* [`src/test/test_dp_base.py`](diffhunk://#diff-8db0beacde4ead52a8da269224e4fbb78b6f7c38d801bf36ccb201ad3f6bcbe1L37-R37): Updated the `namespace` value in test cases to match the schema changes. [[1]](diffhunk://#diff-8db0beacde4ead52a8da269224e4fbb78b6f7c38d801bf36ccb201ad3f6bcbe1L37-R37) [[2]](diffhunk://#diff-8db0beacde4ead52a8da269224e4fbb78b6f7c38d801bf36ccb201ad3f6bcbe1L93-R93)


ahouseholder:
Converted to draft so we don't merge too early

data/schema/v1/Decision_Point-1-0-1.schema.json
```diff
@@ -41,23 +47,32 @@
 		},
 		"namespace": {
 		    "type": "string",
-		    "description": "Namespace (a short, unique string): For example, \"ssvc\" or \"cvss\" to indicate the source of the decision point"
+		    "description": "Namespace (a short, unique string): For example, \"ssvc\" or \"cvss\" to indicate the source of the decision point. See SSVC Documentation for details.",
+		    "pattern": "^[a-z0-9-]{3,4}[a-z0-9/\\.-]*$",
```
tschmidtb51, line 51: In your last approach, one registered value was `cvssv4`. Did you skip the version intentionally?

data/schema/v1/Decision_Point-1-0-1.schema.json
```diff
@@ -41,23 +47,32 @@
...
+		    "examples": ["ssvc", "cvss", "ssvc-jp", "ssvc/acme", "ssvc/example.com"]
 		},
 		"version": {
 		    "type": "string",
-		    "description": "Version (a semantic version string) that identifies this object"
+		    "description": "Version (a semantic version string) that identifies this object",
+		    "pattern": "^(?P<major>0|[1-9][0-9]*)\\.(?P<minor>0|[1-9][0-9]*)\\.(?P<patch>0|[1-9][0-9]*)(?:-(?P<prerelease>(?:0|[1-9][0-9]*|[0-9]*[a-zA-Z-][0-9a-zA-Z-]*)(?:\\.(?:0|[1-9][0-9]*|[0-9]*[a-zA-Z-][0-9a-zA-Z-]*))*))?(?:\\+(?P<buildmetadata>[0-9a-zA-Z-]+(?:\\.[0-9a-zA-Z-]+)*))?$",
```
tschmidtb51, line 57: This is not a valid pattern in JSON as they don't allow named capture groups (see https://semver.org/#is-there-a-suggested-regular-expression-regex-to-check-a-semver-string). You need to use the second suggestion.
```suggestion
	    "pattern": "^(0|[1-9]\d*)\\.(0|[1-9]\d*)\\.(0|[1-9]\\d*)(?:-((?:0|[1-9]\\d*|\\d*[a-zA-Z-][0-9a-zA-Z-]*)(?:\\.(?:0|[1-9]\\d*|\\d*[a-zA-Z-][0-9a-zA-Z-]*))*))?(?:\\+([0-9a-zA-Z-]+(?:\\.[0-9a-zA-Z-]+)*))?$"

**Review by tschmidtb51 — CHANGES_REQUESTED** [review]
Please see my feedback.

File: data/schema/v1/Decision_Point_Group-1-0-1.schema.json [hunk 3]
```diff
@@ -16,16 +16,19 @@
...
-		    "type": "string",
-		    "description": "Version (a semantic version string) that identifies this object"
+		    "$ref": "https://certcc.github.io/SSVC/data/schema/v1/Decision_Point-1-0-1.schema.json#/$defs/decision_point/properties/version"
 		},
 		"name": {
 		    "type": "string",
-		    "description": "A short label that captures the description of the Decision Point or the Group of Decision Points."
+		    "description": "A short label that captures the description of the Decision Point or the Group of Decision Points.",
+		    "minLength": 1,
+		    "examples": ["SSVC Supplier","Coordinator Triage", "SSVC Deployer"]
```
tschmidtb51, line 25: I don't see those in https://github.com/CERTCC/SSVC/tree/main/data/json/decision_points. Shouldn't that be something like `Safety Impact`, `Report Public`, `Technical Impact`?

data/schema/v1/Decision_Point-1-0-1.schema.json
sei-vsarvepalli: Not intentional - but that is a good point. We use the namespace always as `cvss` and never as `cvssv4`  to identify distinct CVSS versions. Discussion needed internally adding @ahouseholder  for awareness.
 
./src/ssvc/decision_points/cvss/base.py 

https://github.com/CERTCC/SSVC/blob/3fd228b2cc83f59e94f7a00bf773233c7671e8c7/src/ssvc/decision_points/cvss/base.py#L28

data/schema/v1/Decision_Point_Group-1-0-1.schema.json
sei-vsarvepalli: These are Decision Point Groups - which targets grouping a set of distinct Decision Points to make sure they are captured according to a `role` - so related Decision Points are bundled together. For e.g., `Technical Impact` is a Decision Point itself , the examples of `SSVC Supplier` is here
https://github.com/CERTCC/SSVC/blob/3fd228b2cc83f59e94f7a00bf773233c7671e8c7/src/ssvc/dp_groups/ssvc/supplier.py#L60

data/schema/v1/Decision_Point-1-0-1.schema.json
```diff
@@ -41,23 +47,32 @@
...
-		    "description": "Namespace (a short, unique string): For example, \"ssvc\" or \"cvss\" to indicate the source of the decision point"
+		    "description": "Namespace (a short, unique string): For example, \"ssvc\" or \"cvss\" to indicate the source of the decision point. See SSVC Documentation for details.",
+		    "pattern": "^[a-z0-9-]{3,4}[a-z0-9/\\.-]*$",
+		    "examples": ["ssvc", "cvss", "ssvc-jp", "ssvc/acme", "ssvc/example.com"]
 		},
 		"version": {
 		    "type": "string",
-		    "description": "Version (a semantic version string) that identifies this object"
+		    "description": "Version (a semantic version string) that identifies this object",
+	    "pattern": "^(0|[1-9]\d*)\\.(0|[1-9]\d*)\\.(0|[1-9]\\d*)(?:-((?:0|[1-9]\\d*|\\d*[a-zA-Z-][0-9a-zA-Z-]*)(?:\\.(?:0|[1-9]\\d*|\\d*[a-zA-Z-][0-9a-zA-Z-]*))*))?(?:\\+([0-9a-zA-Z-]+(?:\\.[0-9a-zA-Z-]+)*))?$"
```
tschmidtb51, line 57: Sorry, I missed to quote two `\`
```suggestion
	    "pattern": "^(0|[1-9]\\d*)\\.(0|[1-9]\\d*)\\.(0|[1-9]\\d*)(?:-((?:0|[1-9]\\d*|\\d*[a-zA-Z-][0-9a-zA-Z-]*)(?:\\.(?:0|[1-9]\\d*|\\d*[a-zA-Z-][0-9a-zA-Z-]*))*))?(?:\\+([0-9a-zA-Z-]+(?:\\.[0-9a-zA-Z-]+)*))?$"

File: data/schema/v1/Decision_Point_Group-1-0-1.schema.json (hunk 3 above)
Comment (tschmidtb51) [reply data/schema/v1/Decision_Point_Group-1-0-1.schema.json:25]: My bad. I didn't realize that this was in the Decision Point Group file.

File: data/schema/v1/Decision_Point-1-0-1.schema.json (hunk 4 above)
Comment (sei-vsarvepalli) [reply data/schema/v1/Decision_Point-1-0-1.schema.json]: comma missing at the end as well that is fixed now in the https://github.com/CERTCC/SSVC/pull/704/commits/0dc08d0092f8a9d49c33ecff5eb9499cf597bf84

File: data/schema/v1/Decision_Point-1-0-1.schema.json [hunk 5]
```diff
@@ -15,23 +15,29 @@
...
+		    "examples": ["One of the following is true: (1) Typical public PoC exists in sources such as Metasploit or websites like ExploitDB; or (2) the vulnerability has a well-known method of exploitation.","Attackers can reliably automate steps 1-4 of the kill chain."]
                 }
             },
 	    "required" : [
 		"key",
 		"name",
 		"description"
 	    ]
-	},
+	},
```
tschmidtb51, line 40:
```suggestion
        },

**Review by tschmidtb51 — COMMENTED** [review]
The line 98in the decision point schema looks off. Shouldn't that be inside a keyword?

File: data/schema/v1/Decision_Point-1-0-1.schema.json (hunk 5 above)
Comment (sei-vsarvepalli) [reply data/schema/v1/Decision_Point-1-0-1.schema.json:40]: I don't see the issue.  May be just indentation that makes it look wrong?

File: data/schema/v1/Decision_Point-1-0-1.schema.json (hunk 1 above)
Comment (ahouseholder) [reply data/schema/v1/Decision_Point-1-0-1.schema.json:51]: At this point, the namespace for any CVSS-derived decision point, regardless of CVSS version, is likely to remain "cvss". We're trying to use the namespace to indicate concepts that are defined by some other authority (e.g., the CVSS SIG in this case) and where our decision points are just intended to reflect their definitions. In other words, we don't "own" the semantics of those decision points, we're just providing them for convenience to SSVC users.

Writing that helps me solidify some thinking about #703 too.

File: data/schema/v1/Decision_Point-1-0-1.schema.json (hunk 4 above)
Comment (ahouseholder) [reply data/schema/v1/Decision_Point-1-0-1.schema.json]: I'd be most comfortable with the second example provided by 

https://semver.org/

specifically

https://regex101.com/r/vkijKf/1/

on the basis that if the semver.org thinks that's the regex, I'd take their word for it.

Comment (sei-vsarvepalli) [reply data/schema/v1/Decision_Point-1-0-1.schema.json]: What we have is exactly that with the double backslashes for usage in the JSON schema that's all.

File: data/schema/v1/Decision_Point-1-0-1.schema.json [hunk 6]
```diff
@@ -41,23 +47,32 @@
...
 		    "type": "string",
-		    "description": "A key (a short, unique string) that can be used to identify the Decision Point/Decision Point value in a shorthand way"
+		    "description": "A key (a short, unique string) that can be used to identify the Decision Point/Decision Point value in a shorthand way",
+		    "minLength": 1,
+		    "examples": ["E", "A"]
 		},
 		"name": {
 		    "type": "string",
-		    "description": "A short label that captures the description of the Decision Point or the Group of Decision Points."
+		    "description": "A short label that captures the description of the Decision Point or the Group of Decision Points.",
```
ahouseholder, line 68: This is the name of the decision point, not the group.

data/schema/v1/Decision_Point-1-0-1.schema.json
```diff
@@ -41,23 +47,32 @@
...
+		    "minLength": 1,
+		    "examples": ["Exploitation", "Automatable"]
 		},
 		"description": {
 		    "type": "string",
-		    "description": "q Description of the Decision Point or the Group of Decision Points as defined."
+		    "description": "Description of the Decision Point or the Group of Decision Points as defined.",
```
ahouseholder, line 74: description of the decision point, not the group

data/schema/v1/Decision_Point_Group-1-0-1.schema.json
```diff
@@ -16,16 +16,19 @@
...
                 },
 		"version": {
-		    "type": "string",
-		    "description": "Version (a semantic version string) that identifies this object"
+		    "$ref": "https://certcc.github.io/SSVC/data/schema/v1/Decision_Point-1-0-1.schema.json#/$defs/decision_point/properties/version"
 		},
 		"name": {
 		    "type": "string",
-		    "description": "A short label that captures the description of the Decision Point or the Group of Decision Points."
+		    "description": "A short label that captures the description of the Decision Point or the Group of Decision Points.",
```
ahouseholder, line 23: this is the name of the decision point group, not the decision point

data/schema/v1/Decision_Point_Group-1-0-1.schema.json
```diff
@@ -16,16 +16,19 @@
...
+		    "minLength": 1,
+		    "examples": ["SSVC Supplier","Coordinator Triage", "SSVC Deployer"]
 		},
 		"description": {
 		    "type": "string",
-		    "description": "Description of the Decision Point or the Group of Decision Points."
+		    "description": "Description of the Decision Point or the Group of Decision Points.",
```
ahouseholder, line 29: description of the decision point group, not decision point

data/schema/v1/Decision_Point-1-0-1.schema.json
tschmidtb51: > Should we stick to [0-9] instead of \d as \d matches say Arabic numerals when regex is used with unicode?

I'd like that as it eliminates the ambiguousness. `\d` is unfortunately sometimes interpreted differently...

data/schema/v1/Decision_Point-1-0-1.schema.json
tschmidtb51: Just a format nit: A trailing space was introduced...

ahouseholder (changes requested):
@sei-vsarvepalli can you please update the description of this PR to say what it actually does (more than linking to the issue)?

I noticed a number of description fields in this changeset that are vague in what they're talking about ("decision point or decision point group", etc.) when in fact they are for one but not the other. I've only commented on the ones that showed up in the changeset, but without looking at the rest of the file I suspect that's an issue throughout the description fields. We should clean that up to avoid confusion.

tschmidtb51:
> The line 98 in the decision point schema looks off. Shouldn't that be inside a keyword?

Sorry - my bad. After I looked at an instance again, I understand now what you are doing.

data/schema/v1/Decision_Point-1-0-1.schema.json
tschmidtb51: Also, if you could fix the indentation, that would help in the review. Many thanks in advance!

data/schema/v1/Decision_Point-1-0-1.schema.json
```diff
@@ -15,23 +15,29 @@
...
                     "type": "string",
-                    "description": "A key (a short, unique string) that can be used to identify the Decision Point/Decision Point value in a shorthand way"
+                    "description": "A key (a short, unique string) that can be used to identify the Decision Point/Decision Point value in a shorthand way",
+		    "minLength": 1,
+		    "examples": ["P", "Y"]
                 },
                 "name": {
                     "type": "string",
-                    "description": "A short label that captures the description of the Decision Point or the Group of Decision Points."
+                    "description": "A short label that captures the description of the Decision Point or the Group of Decision Points.",
```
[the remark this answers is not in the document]
sei-vsarvepalli: Yeh - those have been the same. I have not changed them. Bu happy to change them as well.  You are only seeing it because the "comma" at the end added.

sei-vsarvepalli:
> @sei-vsarvepalli can you please update the description of this PR to say what it actually does (more than linking to the issue)?
> 
> I noticed a number of description fields in this changeset that are vague in what they're talking about ("decision point or decision point group", etc.) when in fact they are for one but not the other. I've only commented on the ones that showed up in the changeset, but without looking at the rest of the file I suspect that's an issue throughout the description fields. We should clean that up to avoid confusion.

All the "descriptions" are from the old JSON file nothing has changed in these. But I do see your point, we can update them too. They were left generic in the first edition, and remained the same so far.

data/schema/v1/Decision_Point-1-0-1.schema.json
sei-vsarvepalli: Accepted this change into a4796e9

ahouseholder:
> > @sei-vsarvepalli can you please update the description of this PR to say what it actually does (more than linking to the issue)?
> > I noticed a number of description fields in this changeset that are vague in what they're talking about ("decision point or decision point group", etc.) when in fact they are for one but not the other. I've only commented on the ones that showed up in the changeset, but without looking at the rest of the file I suspect that's an issue throughout the description fields. We should clean that up to avoid confusion.
> 
> All the "descriptions" are from the old JSON file nothing has changed in these. But I do see your point, we can update them too. They were left generic in the first edition, and remained the same so far.

Yeah, I figured they had been that way for a while. It's up to you, if it's an easy fix this time, I think it's worth doing. If it's not, then we can merge this PR as-is and leave a breadcrumb issue behind to pick up the descriptions at another time.  I don't necessarily want us to block on this PR  just because some description text could be clearer.

Should the changes (outside of the descriptions) imply a versioning event for the SSVC schema? Major, minor or patch, if so?

Does modifying the descriptions imply a versioning event for the SSVC schema? And is it a big deal if we bump versions for stuff like that? The description changes would presumably be like a patch bump rather than a minor or major version I assume.

sei-vsarvepalli:
> > > @sei-vsarvepalli can you please update the description of this PR to say what it actually does (more than linking to the issue)?
> > > I noticed a number of description fields in this changeset that are vague in what they're talking about ("decision point or decision point group", etc.) when in fact they are for one but not the other. I've only commented on the ones that showed up in the changeset, but without looking at the rest of the file I suspect that's an issue throughout the description fields. We should clean that up to avoid confusion.
> > 
> > 
> > All the "descriptions" are from the old JSON file nothing has changed in these. But I do see your point, we can update them too. They were left generic in the first edition, and remained the same so far.
> 
> Yeah, I figured they had been that way for a while. It's up to you, if it's an easy fix this time, I think it's worth doing. If it's not, then we can merge this PR as-is and leave a breadcrumb issue behind to pick up the descriptions at another time. I don't necessarily want us to block on this PR just because some description text could be clearer.
> 
> Should the changes (outside of the descriptions) imply a versioning event for the SSVC schema? Major, minor or patch, if so?
> 
> Does modifying the descriptions imply a versioning event for the SSVC schema? And is it a big deal if we bump versions for stuff like that? The description changes would presumably be like a patch bump rather than a minor or major version I assume.

I would say description change does not require a version bump in the schema (unlike when changing a Decision Point or a Decision Point Group instance). 

The schema version change is really for a few events like
1.  Add/remove elements or change schema in a non-breaking way (addition/patch)
2. Add/remove elements in a potentially-breaking way (revision/minor)
3. Modify the schema in a breaking way checks/formats/required etc. (model/major)

None of these is happening as I see it, so don't see a need to change version number. Adoption of CVE and CSAF program are the only external interactions we are managing, so I think version change is not needed IMO.

If I get time next week I can try to come up with better language for "description" of these fields. This patch mostly is to facilitate a startup adoption by CSAF and CVE programs.



Vijay

sei-vsarvepalli:
This is all complete for review and merge, I added a better description for each JSON schema fields.

ahouseholder (approved):
These all look good. Thanks!

PR #706, opened 2025-02-20
Updated mission_impact.md

sei-bkoo:
Updated wording in light of the fact that SSVC is no longer hosted here as a paper. 

Also removed block of text on Utility that might be causing unnecessary confusion.


sei-bkoo:
After reviewing both Mission Impact and Utility, I think I agree with Allen in that we should just remove that block of text.

ahouseholder (approved):

PR #716, opened 2025-02-24
Add an 'Explanation' Overview document

sei-renae:
The target audience for the Explanation is a non-expert, and per diataxis.fr, should be able to casually read this and get an overview of what is SSVC.

This PR:
Adds the Explanation document
Modifies the main index page cards to Add the Explanation and Delete Learning SSVC
Modifies the menu bar via mkdocs.yml per index page cards
Moves [Learning SSVC](SSVC/tutorials) to Reference

image
image


docs/explanation.md
```diff
@@ -0,0 +1,253 @@
...
+from ssvc.decision_points.system_exposure import LATEST
+from ssvc.doc_helpers import example_block
+
+print(example_block(LATEST, include_json=False))
+```
+
+#### Automability
+
+*Automability* is the same question as contained in the Supplier's Utility decision point.
+For easy reference, *Automability* is a simple yes or no, can the attacker reliably automate all four steps (reconnaissance, weaponization, delivery, exploitation) of the kill chain? If the vulnerability allows remote code execution or command injection, the expected response should be 'yes.'
```
ahouseholder, line 165: again the question mark on `kill chain?`

docs/explanation.md
```diff
@@ -0,0 +1,253 @@
...
+*Mission Impact* is the impact on the Mission Essential Functions of the organization. A **mission essential function (MEF)** is a function “directly related to accomplishing the organization’s mission as set forth in its statutory or executive charter” [@FCD2_2017, page A-1].
+
+```python exec="true" idprefix=""
+from ssvc.decision_points.mission_impact import LATEST
+from ssvc.doc_helpers import example_block
+
+print(example_block(LATEST, include_json=False))
+```
+
+We define the following combinatorics of *Safety Impact* and *Mission Impact* values to determine *Human Impact* values:
```
ahouseholder, line 191: s/combinatorics/combinations/

docs/explanation.md
```diff
@@ -0,0 +1,253 @@
...
+print(example_block(LATEST, include_json=False))
+```
+
+### Coordinator decision points for Triage
+
+<embed src="../pdf/ssvc_2_coord-triage.pdf" alt="Coordination Triage Tree" type="application/pdf"
+style="width: 100%;"
+height = "700" />
+
+The below decision points are  modeled on [CERT/CC Coordinated Vulnerability Disclosure (CVD) project](howto/coordination_triage_decision/#coordinator-triage-units-of-work), but we encourage Coordinators to define their own reasons to engage in Triage.
```
ahouseholder, line 206: CERT/CC's Coordinated Vulnerability Disclosure (CVD) practice.

It's not a project.

ahouseholder, line 206: "reasons to engage in Triage" -- the only reason to engage in triage is when you have to quickly prioritize a large volume of potential tasks for followup.

Also I don't think "triage" is a proper noun.

Now, the _criteria_ for somebody else's triage process might be different. (based on the coordinator's mission and goals)

docs/explanation.md
```diff
@@ -0,0 +1,253 @@
...
+
+Generally speaking, we believe in Supplier agency, and is why the first two questions are about damage control.
```
ahouseholder, line 208: oof. No, that's not right at all.

We're seeking to give priority to cases where our help is actually needed. We don't have the capacity to help everyone with everything. So if the report is already public, there's not much further we can do, except in special circumstances. If the supplier has not been contacted, we'll usually decline in single-supplier cases because we expect the reporter to have taken some effort on their own before coming to us for help.

docs/explanation.md
```diff
@@ -0,0 +1,253 @@
...
+- [Supplier Contacted](reference/decision_points/supplier_contacted.md): If no suppliers have been contacted, then CERT/CC will decline the case unless there are multiple suppliers, [*super effective*](reference/decision_points/system_exposure.md) [Utility](reference/decision_points/utility.md), and [*significant*](reference/decision_points/public_safety_impact.md) [Public Safety Impact](reference/decision_points/public_safety_impact.md).
+  In this case, CERT/CC may encourage the reporter to contact the supplier and submit a new case request if the supplier is unresponsive.
+- [Report Credibility](reference/decision_points/report_credibility.md): If the report is not credible, then CERT/CC will decline the case.
+Please see the [CERT® Guide to Coordinated Vulnerability Disclosure](https://certcc.github.io/CERT-Guide-to-CVD/howto/coordination/_report_credibility/?h=credibilit) for more information about assessing credibility.
```
ahouseholder, line 214: drop the "?h=credibility" from the url. We don't need search highlights in the linked page.

docs/explanation.md
```diff
@@ -0,0 +1,253 @@
...
+- [Supplier Cardinality](reference/decision_points/supplier_cardinality.md): Cases involving multiple suppliers can get complicated very quickly, so we are more likely to get involved in those cases.
+- [Supplier Engagement](reference/decision_points/supplier_engagement.md): If the suppliers are already engaged in a case, there is usually less for a coordinator to do, making it less likely that we will coordinate a case.
+- [Utility](reference/decision_points/utility.md): If the vulnerability has high utility, then CERT/CC is more likely to coordinate the case.
+- [Public Safety Impact](reference/decision_points/public_safety_impact.md): If the vulnerability has significant
+   public safety impact, then CERT/CC is more likely to coordinate the case.
+
+The last two questions, *Utility* and *Public Safety Impact*, are the same as asked in the Supplier tree.
```
ahouseholder, line 221: s/tree/decision model/

docs/explanation.md
```diff
@@ -0,0 +1,253 @@
...
+
+### Coordinator decision points for Publication
+
+<embed src="../pdf/ssvc_2_coord-publish.pdf" alt="Suggested tree for a coordinator's publication decision" type="application/pdf"
+style="width: 100%;"
+height = "600" />
+
+A decision to publish is determined after criteria for Publication are met. SSVC adds value by codifying publication criteria so that the decision is explainable. Publishing a vulnerability should add public value, and a Coordinator must decide why they should publish instead of or in addition to the Supplier. Again, the below decision points are modeled on the [CERT/CC Coordinated Vulnerability Disclosure (CVD) project](howto/coordination_triage_decision/#coordinator-triage-units-of-work), but we encourage Coordinators to define their own rationales for publication.
```
ahouseholder, line 229: At the CERT/CC, a decision to publish...

ahouseholder, line 229: > and a Coordinator must decide why they should publish instead of or in addition to the Supplier

Again, _we_ do this, and we're providing it as an example. It's not a requirement or expectation of Coordinators in general to make this decision using the same criteria.

Coordinators don't have to publish at all if that's not part of their mission.

docs/explanation.md
```diff
@@ -0,0 +1,253 @@
...
+
+The publication decision reuses the [*Exploitation*](reference/decision_points/exploitation.md) decision point and adds two new ones ([*Supplier Involvement*](reference/decision_points/supplier_involvement.md) and [*Public Value Added*](reference/decision_points/public_value_added.md)).
+
+- [*Supplier Involvement*](reference/decision_points/supplier_involvement.md) - If the supplier is involved and likely to publish already, there is less need for the CERT/CC to publish.
```
ahouseholder, line 233: While too technical to go into in this document, see https://certcc.github.io/Vultron/reference/ssvc_crosswalk/#supplier-involvement for a brief "Engagement vs. Involvement: What's the Difference?" Might be able to summarize that distinction here.

docs/explanation.md
```diff
@@ -0,0 +1,253 @@
...
+- [*Exploitation*](reference/decision_points/exploitation.md) - If the vulnerability is being actively exploited, the CERT/CC is more likely to publish. Exploitation has the same values as Supplier Exploitability.
```
ahouseholder, line 234: _Supplier Exploitability_ is not a thing.

It's _Exploitation_, and it has the same meaning wherever it appears.

docs/explanation.md
```diff
@@ -0,0 +1,253 @@
...
+- [*Public Value Added*](reference/decision_points/public_value_added.md) - If there is already significant public discussion of the vulnerability, there might not be much for the CERT/CC to add, making us less likely to publish.
+
+## Final notes
+
+SSVC enables different stakeholders to make determinations about how a vulnerability will impact them, how to respond to a vulnerability, and how to communicate and explain their decisions. Unlike existing vulnerability scoring systems, SSVC evaluates risk to a particular stakeholder without imposing a one-size-fits-all approach. The below subsections describe some other details about SSVC's features and provide links to further reading about SSVC and CVD.
```
ahouseholder, line 239: s/existing/other/

docs/explanation.md
```diff
@@ -0,0 +1,253 @@
...
+
+### Multiple vulnerabilities of the same priority
+
+Naturally, a stakeholder will have multiple vulnerabilities at any given time, and it is likely that there will be multiple vulnerabilities of the same priority. The stakeholder can choose how to order vulnerabilities within the same priority- vulnerabilities of the same priority can be addressed in *any* order.
+
+### Communicating results
+
+Coordinators will be especially concerned with communicating information about vulnerabilities, but Suppliers will be just as concerned about communicating to Deployers. Because information changes over time, SSVC decisions should always be timestamped and marked with the version of SSVC that was used. We provide a self-contained, machine-readable [JSON schema]() and an abbreviated ["vector string" notation](). Please note that while the JSON schema can be compressed to an abbreviated vector string, we do not currently support converting from an abbreviated vector string to the full JSON schema.
```
ahouseholder, line 247: We want to move away from the vector string notation, so I would suggest not mentioning it here.

ahouseholder (changes requested):
A few general thoughts:

- I'm concerned about length for an intro doc
- It might be better to give readers the intro, then let them do a choose-your-own-adventure based on what roles they have. I'm thinking something like:

```mermaid
flowchart TB

intro --> stakeholders
stakeholders --> supplier
stakeholders --> deployer
stakeholders --> coordinator
supplier --> sddp[supplier decisions and decision points]
deployer --> dddp[deployer decisions and decision points]
coordinator --> cddp[coordinator decisions and decision points]
sddp --> fn[final notes]
dddp --> fn
cddp --> fn
```

This could be done as a series of pages (one for each stakeholder role) that includes
links to each of the other paths.

If you want to keep it in one page, we could do it using [content tabs](https://squidfunk.github.io/mkdocs-material/reference/content-tabs/) That might get weird with longer content though.

docs/explanation.md
```diff
@@ -0,0 +1,253 @@
+# What is SSVC
+
+## Intro
+
+Risk exists on a continuum, and different stakeholders have different threat models and risk appetites.
```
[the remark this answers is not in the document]
sei-renae: Due to the audience of the document, if I were to change it, I would rather append: "Cybersecurity risk is no exception."

docs/explanation.md
```diff
@@ -0,0 +1,253 @@
...
+
+Take, for example, an exported software library that contains many functions, one of which is vulnerable to SQL injection. Your software imports this library but does not use this vulnerable function. Is your software, in its current state, at risk of this vulnerability? You likely answered, "probably not."
+In this example, you are the *Deployer* of software that contains a vulnerability, but this vulnerability is not very risky to your business, so you are not very concerned. Conversely, the exported software library's owner is the *Supplier*, and they probably want the vulnerability patched as quickly as possible.
+
+Unlike other vulnerability categorization systems that rate on technical severity (how bad it could be if the vulnerability were exploited) or exploitability (how likely it is that there will be an exploit), SSVC rates vulnerabilities on *risk* to the concerned stakeholder. It is not an attempted one-size-fits-all solution. SSVC follows a stakeholder-defined algorithm to make these determinations of risk.
```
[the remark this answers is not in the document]
sei-renae: > (how bad it could be if the vulnerability were exploited)
is my definition of 'technical severity'

> (how likely it is that there will be an exploit)
I paraphrased from "The Exploit Prediction Scoring System (EPSS) is 'a data-driven effort for estimating the likelihood (probability) that a software vulnerability will be exploited in the wild.'" (SSVC whitepaper v2 p68)

ack on 'algorithm'

docs/explanation.md
sei-renae: Triage is capitalized because of title case (like I do with Suppliers, Deployers, and Coordinators)

docs/explanation.md
sei-renae: In trying to keep this agnostic from CERT/CC's Coordinator decision model, I suggest:

> ## "What is the difference between *Supplier Engagement* and *Supplier Involvement*?"
> 
>     *Supplier Engagement*, asks whether a Supplier has acknowledged the vulnerability and is participating in the coordination effort.
>     *Supplier Involvement*, asks what, if anything, the Supplier is doing about the vulnerability.

docs/explanation.md
```diff
@@ -0,0 +1,253 @@
+# What is SSVC
+
+## Intro
+
+Risk exists on a continuum, and different stakeholders have different threat models and risk appetites.
+
+Take, for example, an exported software library that contains many functions, one of which is vulnerable to SQL injection. Your software imports this library but does not use this vulnerable function. Is your software, in its current state, at risk of this vulnerability? You likely answered, "probably not."
```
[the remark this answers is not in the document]
sei-renae: How about:
_Imagine you're a software developer, and you just found out about a vulnerability in a library upon which your product depends. You are now concerned that your software is also vulnerable. Upon investigation, you learn that your software does not use the vulnerable part of the imported library. Is your software, in its current state, at risk of this vulnerability? You likely answered, "probably not."_

_In this example, you are a Supplier of software that has an upstream Supplier vulnerability, but this vulnerability is not very risky to your business, so you are not very concerned. Conversely, the upstream Supplier probably wants the vulnerability patched as quickly as possible._

docs/explanation.md
```diff
@@ -0,0 +1,268 @@
...
+
+In this example, you are a Supplier of software that has an upstream Supplier vulnerability, but this vulnerability is not very risky to your business, so you are not very concerned. Conversely, the upstream Supplier probably wants the vulnerability patched as quickly as possible.
+
+Unlike other vulnerability categorization systems that rate on technical severity (how bad it could be if the vulnerability were exploited) or exploitability (how likely it is that there will be an exploit), SSVC rates vulnerabilities on *risk* to the concerned stakeholder. It is not an attempted one-size-fits-all solution. SSVC follows a stakeholder-defined decision model to make these determinations of risk.
+
+!!! note "Caveats"
+
+    All information in this document is suggested following research, but we must reiterate: it is *not a one-size-fits-all* system. We encourage stakeholders to define to define their own decision points, timeframes of actionability, and possibly, even their own roles.
```
sei-bkoo, line 15: Typo "to define to define".

sei-bkoo (commented):
Here are some thoughts:

- I would consider removing the code blocks that appear after each decision point definition. A first time reader might be confused for what these are for (especially those who might stumble upon this on the internet). Apologies if they don't appear or if there is a good reason for those to be there.
- Something to consider down the line (which I actually want to work on) is some kind of example (text or visual) to keep enhancing accessibility. See examples [here](https://rogerdudler.github.io/git-guide/) and [here](https://devhints.io/git-branch). Whether it should be on the same page as this somehow or as another page, we can discuss later!

sei-renae:
Also at https://github.com/sei-renae/SSVC/discussions/1

docs/explanation.md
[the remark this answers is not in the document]
ahouseholder: Yeah that makes sense.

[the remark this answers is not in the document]
sei-renae: This is the only change I made other than the markdownlint

ahouseholder (changes requested):
1. Link Checker is failing on a few links (see checks below)
2. One last change request: could you please rename `explanation.md` to either 

- `ssvc_overview.md` (matches the nav menu name)
- `what_is_ssvc.md` (matches the h1 in the page itself)

I'm leaning toward `ssvc_overview.md` since the nav is less likely to change in the future than the page title.

This is entirely an "ease of future maintenance" request, so you don't have to consult the `mkdocs.yml` or the home page to figure out which file contains the SSVC overview doc. `explanation.md` is too generic to be obvious.

ahouseholder:
Still need to fix these:

```
URL        `topics/decision_trees/'
Name       `Decision Trees'
Parent URL file:///home/runner/work/SSVC/SSVC/site/ssvc_overview/index.html, line 3943, col 314
Real URL   file:///home/runner/work/SSVC/SSVC/site/ssvc_overview/topics/decision_trees/
Check time 0.000 seconds
Result     Error: URLError: <urlopen error [Errno 2] No such file or directory: '/home/runner/work/SSVC/SSVC/site/ssvc_overview/topics/decision_trees/'>

URL        `howto/coordination_triage_decision/#coordinator-triage-units-of-work'
Name       `CERT/CC Coordinated Vulnerability Disclosure (CVD) project'
Parent URL file:///home/runner/work/SSVC/SSVC/site/ssvc_overview/index.html, line 3954, col 48
Real URL   file:///home/runner/work/SSVC/SSVC/site/ssvc_overview/howto/coordination_triage_decision/
Check time 0.000 seconds
Result     Error: URLError: <urlopen error [Errno 2] No such file or directory: '/home/runner/work/SSVC/SSVC/site/ssvc_overview/howto/coordination_triage_decision/'>
10 threads active,    [37](https://github.com/CERTCC/SSVC/actions/runs/15495231228/job/43630238573?pr=716#step:7:38) links queued,  795 links in 842 URLs checked, runtime 11 seconds

URL        `reference/index'
Name       `Reference material'
Parent URL file:///home/runner/work/SSVC/SSVC/site/ssvc_overview/index.html, line 4220, col 4
Real URL   file:///home/runner/work/SSVC/SSVC/site/ssvc_overview/reference/index
Check time 0.000 seconds
Result     Error: URLError: <urlopen error [Errno 2] No such file or directory: '/home/runner/work/SSVC/SSVC/site/ssvc_overview/reference/index'>

URL        `howto/tree_customization'
Name       `Guidance for customizing a Decision Tree'
Parent URL file:///home/runner/work/SSVC/SSVC/site/ssvc_overview/index.html, line [42](https://github.com/CERTCC/SSVC/actions/runs/15495231228/job/43630238573?pr=716#step:7:43)21, col 4
Real URL   file:///home/runner/work/SSVC/SSVC/site/ssvc_overview/howto/tree_customization
Check time 0.000 seconds
Result     Error: URLError: <urlopen error [Errno 2] No such file or directory: '/home/runner/work/SSVC/SSVC/site/ssvc_overview/howto/tree_customization'>
```

ahouseholder (approved):

PR #727, opened 2025-02-27
Update json examples in repository

ahouseholder:
Forgot to run doctools.py after we modified some of the decision points. This only affects the examples in the `/data/json/decision_points` dir in the repository, the site now generates the reference examples on the fly because of #683


sei-vsarvepalli (changes requested):
Can we add an index.json or index.md that lists all the files in each of these folders
`data/json/` recursively ? I am happy to code this if we think it is needed.

ahouseholder:
> Can we add an index.json or index.md that lists all the files in each of these folders `data/json/` recursively ? I am happy to code this if we think it is needed.

I'm not against doing us doing this, but I would rather not have that block this PR. There are a lot of things in `data/` that are _not_ generated by `doctools.py` so it'd be more than just augmenting the loop that dumps the `*.json` files.

Captured the request in #728

sei-vsarvepalli (approved):

PR #760, opened 2025-03-26
Update Publish branch with the latest updates to ensure SSVC newer version of schema is available.

sei-vsarvepalli:
This is related to OASIS TC request for CSAF taking in SSVC https://github.com/oasis-tcs/csaf/issues/803 to ensure we have the ability to provide the latest schema files over the publish branch. We can fork a different branch if needed to move publish very specific to these files. It looks like none of the changes would cause problems from testing in my fork at least.

Vijay


ahouseholder (approved):
Tagged release v2025.3 to this PR commit.

PR #795, opened 2025-06-23
Initial Decision Table object

### LINKED ISSUE #592: We need a Policy object

The PolicyGenerator object can be used to generate a CSV formatted policy, but we should really have a separate object to represent a policy.

See also #591

ahouseholder:
- resolves #592 

This PR creates a `DecisionTable` object that contains:

- a list of decision points to be used as inputs
- an outcome group (technically another decision point object) to be used as output
- a mapping of combinations of decision point values to outcome values

While the `DataTable` object can generate a CSV file as output, importing of CSV tables to populate a `DecisionTable` object is not supported in this implementation.

This PR is just intended to provide the base data structure because other work depends on the existence of this object.

Future work not addressed in this PR:

- Integrate the `DecisionTable` object with the `PolicyGenerator` tool, which does things like validates a decision table/policy for logical consistency with the partial ordering of decision point values and outcome value combinations. Ideally, we'd want a DecisionTable object to be able to validate its internal structure accordingly. Some starting attempts are found in the `experimental_base.py` module, but these are not fully integrated and will need additional work before they can be rolled into the main module.
- Bring the generated JSON in line with schemas (or modify schemas to match this object)
- Tools to create, explore, and edit `DecisionTable` objects
- Create a data class to represent a selection from a `DataTable` object. This might look like a Mapping object with a downselection of rows. 
- Support import of CSV format to reconstitute a `DecisionTable` object

I plan to create separate issues for each of these once this PR is merged.

## CoPilot Summary

This pull request introduces significant updates to the SSVC (Stakeholder-Specific Vulnerability Categorization) decision table framework. The changes include the addition of new decision table models, enhancements to the mapping and serialization logic, and updates to the build process. Below is a summary of the most important changes grouped by theme.

### New Decision Table Models and Features:
* Added a new `DecisionTable` class in `src/ssvc/decision_tables/base.py` with functionality to generate, validate, and export decision table mappings. This includes methods for creating full mappings, distributing outcomes, and exporting mappings to CSV.
* Introduced an experimental `DecisionTable` class in `src/ssvc/decision_tables/experimental_base.py` with advanced features like consistency checks, graph-based validation, and mapping serialization to CSV.

### Licensing and Documentation:
* Added detailed copyright and licensing headers to `src/ssvc/decision_tables/__init__.py`, `src/ssvc/decision_tables/base.py`, and `src/ssvc/decision_tables/experimental_base.py`. These headers outline the terms of use and distribution for the code. [[1]](diffhunk://#diff-4ddad1c3acdc06bdf71b689776d2144f44d855da7b5676124f99dfcd4d42725dR1-R18) [[2]](diffhunk://#diff-17e72e9c09ff4865ce37c3185017aafb42a8a15b4d99b08674b6279b29f5e4c9R1-R245) [[3]](diffhunk://#diff-7006177f273a858b7b240fd475dfb028ebb47d26762c1a7eb6a7a358a0dfde00R1-R399)

### Build Process Enhancements:
* Updated the `Makefile` to include a step for building the latest Docker test image before running tests. This ensures the test environment is up-to-date.


ahouseholder:
> I think the Decision TableJSON schema `mappings` field should match the `selections` in ../data/schema/current/Decision_Point_Value_Selection.schema.json - so basically people should be able go from selections in the a DP Value Schema and iterate to find all the possible outcomes.

If I'm interpreting the schema right, I don't think this will work.

1. a selection object  is basically just a DecisionPoint object with a (potentially) reduced set of values:

```python
class SsvcDecisionPointSelection(BaseModel):
    name: str = Field(..., min_length=1, description="Name of the decision point")
    namespace: str = Field(
        ..., min_length=1, description="Namespace of the decision point"
    )
    version: str = Field(..., min_length=1, description="Version of the decision point")
    values: List[str] = Field(
        ..., min_items=1, description="Selected values for the decision point"
    )
```

This doesn't fit as a mapping because it doesn't reflect the specific mapping of `((d1v1, d2v1, d3v1), o1v1)` etc. For example, https://github.com/CERTCC/SSVC/blob/main/data/schema_examples/CVE-1900-1234-Decision_Point_Value_Selection.json does not even mention an outcome value. So this object cannot be used to describe a policy mapping.

2. A decision point value selection object:

```python
class DecisionPointValueSelection(BaseModel):
    id: str = Field(
        ...,
        min_length=1,
        description="Identifier for the vulnerability (e.g., CVE, VU#, GHSA, etc.)",
    )
    role: str = Field(
        ...,
        min_length=1,
        description="Stakeholder role (Supplier, Deployer, Coordinator, etc.)",
    )
    timestamp: datetime = Field(
        ..., description="RFC 3339 timestamp for the evaluation"
    )
    schemaVersion: str = Field(
        ...,
        min_length=1,
        description="Schema version used for the decision point evaluation",
    )
    selections: List[SsvcDecisionPointSelection] = Field(
        ..., min_items=1, description="Evaluated decision points"
    )
```

However, the fields `id`, `role`, `timestamp`, and `schema version` are out of place for a decision policy, as they are specific to a particular evaluation of a specific vul at some point in time by a given role. The only thing left here is the "selections", but see item 1 above for why that's not going to work.

I'm also suggesting we punt the "How to represent selection" problem to

- #800

> I think the CSV is also very noisy, the column headers have say `ssvc:SI:1.0.0`, each row also has the same column name is repeated. As the Column is known already the header, there is no reason to repeat it `ssvc:SI:1.0.0:FR`

Agreed. I will look into maybe passing this through a Pandas dataframe before we emit the CSV so we can make it a little less chatty.

sei-vsarvepalli:
> > I think the Decision TableJSON schema `mappings` field should match the `selections` in ../data/schema/current/Decision_Point_Value_Selection.schema.json - so basically people should be able go from selections in the a DP Value Schema and iterate to find all the possible outcomes.
> 
> If I'm interpreting the schema right, I don't think this will work.
> 
> 1. a selection object  is basically just a DecisionPoint object with a (potentially) reduced set of values:
> 
> ```python
> class SsvcDecisionPointSelection(BaseModel):
>     name: str = Field(..., min_length=1, description="Name of the decision point")
>     namespace: str = Field(
>         ..., min_length=1, description="Namespace of the decision point"
>     )
>     version: str = Field(..., min_length=1, description="Version of the decision point")
>     values: List[str] = Field(
>         ..., min_items=1, description="Selected values for the decision point"
>     )
> ```
> 
> This doesn't fit as a mapping because it doesn't reflect the specific mapping of `((d1v1, d2v1, d3v1), o1v1)` etc. For example, https://github.com/CERTCC/SSVC/blob/main/data/schema_examples/CVE-1900-1234-Decision_Point_Value_Selection.json does not even mention an outcome value. So this object cannot be used to describe a policy mapping.
> 
> 2. A decision point value selection object:
> 
> ```python
> class DecisionPointValueSelection(BaseModel):
>     id: str = Field(
>         ...,
>         min_length=1,
>         description="Identifier for the vulnerability (e.g., CVE, VU#, GHSA, etc.)",
>     )
>     role: str = Field(
>         ...,
>         min_length=1,
>         description="Stakeholder role (Supplier, Deployer, Coordinator, etc.)",
>     )
>     timestamp: datetime = Field(
>         ..., description="RFC 3339 timestamp for the evaluation"
>     )
>     schemaVersion: str = Field(
>         ...,
>         min_length=1,
>         description="Schema version used for the decision point evaluation",
>     )
>     selections: List[SsvcDecisionPointSelection] = Field(
>         ..., min_items=1, description="Evaluated decision points"
>     )
> ```
> 
> However, the fields `id`, `role`, `timestamp`, and `schema version` are out of place for a decision policy, as they are specific to a particular evaluation of a specific vul at some point in time by a given role. The only thing left here is the "selections", but see item 1 above for why that's not going to work.
> 
> I'm also suggesting we punt the "How to represent selection" problem to
> 
> * [Create dataclass to represent DecisionTable selection #800](https://github.com/CERTCC/SSVC/issues/800)
> 
> > I think the CSV is also very noisy, the column headers have say `ssvc:SI:1.0.0`, each row also has the same column name is repeated. As the Column is known already the header, there is no reason to repeat it `ssvc:SI:1.0.0:FR`
> 
> Agreed. I will look into maybe passing this through a Pandas dataframe before we emit the CSV so we can make it a little less chatty.

To be clear I mean the Decision Table JSON format will be hard to punt it off to use #800 in my mind. I think if we can basically mimic  the ../data/schema/current/Decision_Point_Value_Selection.schema.json#/$defs/SsvcdecisionpointselectionSchema section of the schema for this Class it will help other things fall in place including the CSV. 

So for example for Deployer decision table the very first row of CSV will look like this (basically matches .  While it is noisy it will capture the each column of the row in "gory" details - so one can capture the matching row-column combo to find all the DSOI matching values in this example. 

```
            {
                "namespace": "ssvc",
                "version": "1.0.0",
                "values": [
                    "none"
                ],
                "name": "Exploitation"
            },
            {
                "namespace": "ssvc",
                "version": "1.0.1",
                "values": [
                    "small"
                ],
                "name": "Exposure"
            },
            {
                "namespace": "ssvc",
                "version": "2.0.0",
                "values": [
                    "no"
                ],
                "name": "Automatable"
            },
            {
                "namespace": "ssvc",
                "version": "1.0.0",
                "values": [
                    "low"
                ],
                "name": "Human Impact"
            },
            {
                "namespace": "ssvc",
                "version": "1.0.0",
                "values": [
                    "defer"
                ],
                "name": "Defer, Scheduled, Out-of-Cycle, Immediate"
            }
            
```


This will be the first row in the older CSV deployer-options.csv 

`Exploitation,Exposure,Automatable,Human Impact,Priority
none,small,no,low,defer`

or the newer CSV

`ssvc:E:1.0.0,ssvc:EXP:1.0.0,ssvc:A:2.0.0,ssvc:HI:2.0.1,ssvc:DSOI:1.0.0
N,S,N,L,D
`

Hope that makes sense.

sei-vsarvepalli:
I am less concerned about aesthetic but more concerned about problems like parsing and structuring strictly using a delimiter and an array order (first element is namespace etc..). We should avoid any such mapping as it will lead to trouble in the long run. If we want ssvc:SINV:1.0.0 to be unique and reused, we could use local short-form (or long-from) UID's for decision points, instead of parsing to determine the meaning, it will become a reliable look-up within the document. So the decision points can become a dictionary with UIDs or some unique ids like

```json
{"decision_points": {
 "7504-0E04" : {
        "name": "Exploitation",
        "description": "The present state of exploitation of the vulnerability.",
        "namespace": "ssvc",
        "version": "1.0.0",
        "schemaVersion": "1-0-1",
        "key": "E",
         "values": [
...
        ]
 }
 "9309-1104" : {
       "name": "Exploitability",
       "description": "This metric measures the current state of exploit technique or code availability and suggests a likelihood of exploitation.",
       "namespace": "cvss",
       "version": "1.1.0",
       "schemaVersion": "1-0-1",
       "key": "E",
         "values": [
...
        ]
  "1201-010E": {
    "name": "MoSCoW",
    "description": "The MoSCoW (Must, Should, Could, Won't) outcome group.",
    "namespace": "x_basic",
    "version": "1.0.0",
    "schemaVersion": "1-0-1",
    "key": "MSCW",
     "values": [
...
     ]
 }
}
```

There is no reason for the outcome to be distinct from a decision point on the schema as well , as outcomes are also essentially decision points and could be input for other decision tables/evaluations. Our "compound decision points" provides some of this recursion so decision tables can be recursed as well.  The decision table can also be a lookup that goes from one definition of "Exploitation" (SSVC) to another "Exploitability" (in CVSS) can also be represent in the mapping. So the mapping in that case will look something like below:

```json

{"mapping" [
   {
     "7504-0E04": "N",
     "9399-0D0A": "FR",
     "2301-2011": "L",
     "1201-010E": "W" 
 }
 {
    "7504-0E04": "N",
    "9309-1104": "U"
 }
]}
```

ahouseholder:
I don't really like the idea of introducing another ID when the `ns:key:version` is already an effective identity for a decision point. How about if we target something like this:

```json

{"decision_points": {
  "ssvc:E:1.0.0": {
    "name": "Exploitation",
    "description": "The present state of exploitation of the vulnerability.",
    "namespace": "ssvc",
    "version": "1.0.0",
    "schemaVersion": "1-0-1",
    "key": "E",
    "values": [
      {
        "key": "N",
        "name": "None",
        "description": "There is no evidence of active exploitation and no public proof of concept (PoC) of how to exploit the vulnerability."
      },
...
    ]
  },
  "ssvc:EXP:1.0.0": {
    "name": "System Exposure",
    "description": "The Accessible Attack Surface of the Affected System or Service",
    ...
"mapping": [
  {
    "ssvc:E:1.0.0": "N",
    "ssvc:EXP:1.0.0": "S",
    "ssvc:MI:1.0.0": "N",
    "ssvc:SI:1.0.0": "N",
    "x_basic:MSCW:1.0.0": "W"
  },
  {
    "ssvc:E:1.0.0": "N",
    "ssvc:EXP:1.0.0": "S",
    "ssvc:MI:1.0.0": "N",
    "ssvc:SI:1.0.0": "M"
    "x_basic:MSCW:1.0.0": "W"
  },
  ...
]
```
If we're going with dicts though, we need a way to discriminate the outcome decision point from the input ones. I'd prefer to keep that as `{"decision_points": {...}, "outcome": {...}, "mapping": []}` where the keys for "decision_points" and "outcome" are just the `dp.id` as shown above, and the keys of the dicts in each row of "mapping" are the same. That avoids having to parse the colon-separated keys aside from just treating them as lookup keys for a dict.

ahouseholder:
## Where (I think) we agree

I think we're effectively converging on a data structure that will look like:

```python
{ "decision_points": { 
    "k1": { DecisionPoint1 },
    "k2": { DecisionPoint2 },
    ...
    "kN": { DecisionPointN},
    },
    "outcome": "kN",
    "mapping": [
         {"k1":"k1v1", "k2":"k2v1",..."kN":"kNv1"},
         ...,
         {"k1":"k1vX", "k2":"k2vY",...,"kN":"kNvZ"}
     ]
    }
```

Where:
`decision_points` is a dict of decision points.
`outcome` is just the key of the decision point to be used as the outcome group (i.e., a pointer back to the `decision_points` dict)
`mapping` is a list of dicts where each row has the same keys as the `decision_points` dict, but the values are just the `value.key`.

### Agreement Check

Do you agree so far? If not, we should figure out the structural problem before proceeding to the next part.


## Disagreement: What keys to use for the dicts?

We currently lack agreement on how to derive `k1`, `k2`,...,`kN`:

- You've suggested a hex ID string that I don't know how to generate, even after looking at the [CSAF Product ID](https://docs.oasis-open.org/csaf/csaf/v2.0/os/csaf-v2.0-os.html#3132-full-product-name-type---product-id) documentation. (I just need more info on how those `ABCD-EFGH`-style IDs are generated. Presumably some sort of hash function, but they seem really short for most strong hashes, with only 4 bytes of output.)
-  However, the whole *Product ID* thing seems like unnecessary complexity to me here, because whereas identifying products is known to be hard, in our case, we just need an identity function for each decision point, which we have (see next point)
- We've established that the tuple `(namespace, decision_point_key, decision_point_version)` is intended to be a globally unique identity for any Decision Point.
- But tuples are clumsy to use as dict keys. So I proposed converting them to a colon-separated string, essentially: `ns:key:ver`
- Now if we were to use a hash function to generate an `ABCD-EFGH` style key, what would we use as input to that? I'd expect it'd be the `ns:key:ver` string, because that's already a known identity for the decision point.
- Which makes it seem to me that the only reason to hash it would be to obfuscate it, and I'm not seeing why that would be a good idea.

To summarize, the contention seems to be between:

`kn = "ns:key:ver"` vs `kn = hash("ns:key:ver")`

for some yet-to-be-named `hash()`

and I'm just not getting why we'd want the hashed string when we're not trying to hide the string.

## Implications

This probably implies a change to the `DecisionPointGroup` object as well, to use the `dict[str,DecisionPoint]` instead of the current `tuple[DecisionPoint,None,None]`, which means that the Decision Point Group json schema will need to be updated too. (But we need to settle this out first.)

ahouseholder:
A few quick notes before I get to the main point of this comment:

---

_Stakeholders are not strongly associated to Groups_

> for now we have identified a Stakeholder to be Group.

I don't think we have. We have groups for some specific legacy decision trees, but also groups for SSVCv1, SSVCv2.1, CVSSv3, etc. There is no strong association that Group = Stakeholder. It's just a container for any set of decision points you want to put into a group.

---

_Schema changes to DecisionPointGroup coming soon_

The unit tests are currently failing because changes to the DecisionPointGroup object broke the schema validation. I'm planning to work on fixing that next. The schema is going to have to change, and I will probably take the opportunity to pick up #734 while we're at it.

---

_If we need to represent multiple trees in a single json file, that's a **later** problem_

I see where you're going with the idea about having compound decisions in a single object. But that would be a complication of the current object, and I think we can get by with the simplified version for now. As it stands, we can make convenient assumptions like:

- every item in the decision point group should be represented in every row in the mapping
- there is only one outcome and it's just another element in the decision point group

If we go toward compound decisions, or even "list of decision models" without interdependence, this would make more sense to me to be a new object that just has a list of DecisionTables in it. For example, "Coordinator Decisions" might be a simple JSON list including two independent DecisionTables, one for _triage_ and one for _publication_, based on our existing models. 

We can spawn a new issue for that if needed.

---

# Current DecisionTable implementation status

Okay, on to the main event. 

I have implemented both the non-obfuscated `ns:key:ver` style and the obfuscated 4-byte random hex string in `NNNN-NNNN` format. I'm really struggling to see why the obfuscated version would be better. Making them unique each time you generate the JSON seems odd to me, but that's how I implemented the `.obfuscate()` methods.

Extended details below the twisties...

<details>

<summary>JSON DecisionTable without obfuscation</summary>

## Without Obfuscation

```python
dt = DecisionTable(...)
dt.model_dump_json(indent=2)
```
results in

```json

{
  "name": "Test Table",
  "description": "A test decision table",
  "namespace": "x_test",
  "version": "0.0.0",
  "schemaVersion": "1-0-1",
  "decision_point_group": {
    "version": "1.0.0",
    "schemaVersion": "1-0-1",
    "name": "Coordinator Publication",
    "description": "The decision points used by the coordinator during publication.",
    "decision_points": {
      "ssvc:SINV:1.0.0": {
        "name": "Supplier Involvement",
        "description": "What is the state of the supplier’s work on addressing the vulnerability?",
        "namespace": "ssvc",
        "version": "1.0.0",
        "schemaVersion": "1-0-1",
        "key": "SINV",
        "values": [
          {
            "key": "FR",
            "name": "Fix Ready",
            "description": "The supplier has provided a patch or fix."
          },
          {
            "key": "C",
            "name": "Cooperative",
            "description": "The supplier is actively generating a patch or fix; they may or may not have provided a mitigation or work-around in the mean time."
          },
          {
            "key": "UU",
            "name": "Uncooperative/Unresponsive",
            "description": "The supplier has not responded, declined to generate a remediation, or no longer exists."
          }
        ]
      },
      "ssvc:E:1.0.0": {
        "name": "Exploitation",
        "description": "The present state of exploitation of the vulnerability.",
        "namespace": "ssvc",
        "version": "1.0.0",
        "schemaVersion": "1-0-1",
        "key": "E",
        "values": [
          {
            "key": "N",
            "name": "None",
            "description": "There is no evidence of active exploitation and no public proof of concept (PoC) of how to exploit the vulnerability."
          },
          {
            "key": "P",
            "name": "PoC",
            "description": "One of the following cases is true: (1) private evidence of exploitation is attested but not shared; (2) widespread hearsay attests to exploitation; (3) typical public PoC in places such as Metasploit or ExploitDB; or (4) the vulnerability has a well-known method of exploitation."
          },
          {
            "key": "A",
            "name": "Active",
            "description": "Shared, observable, reliable evidence that the exploit is being used in the wild by real attackers; there is credible public reporting."
          }
        ]
      },
      "ssvc:PVA:1.0.0": {
        "name": "Public Value Added",
        "description": "How much value would a publication from the coordinator benefit the broader community?",
        "namespace": "ssvc",
        "version": "1.0.0",
        "schemaVersion": "1-0-1",
        "key": "PVA",
        "values": [
          {
            "key": "L",
            "name": "Limited",
            "description": "Minimal value added to the existing public information because existing information is already high quality and in multiple outlets."
          },
          {
            "key": "A",
            "name": "Ampliative",
            "description": "Amplifies and/or augments the existing public information about the vulnerability, for example, adds additional detail, addresses or corrects errors in other public information, draws further attention to the vulnerability, etc."
          },
          {
            "key": "P",
            "name": "Precedence",
            "description": "The publication would be the first publicly available, or be coincident with the first publicly available."
          }
        ]
      },
      "x_basic:MSCW:1.0.0": {
        "name": "MoSCoW",
        "description": "The MoSCoW (Must, Should, Could, Won't) outcome group.",
        "namespace": "x_basic",
        "version": "1.0.0",
        "schemaVersion": "1-0-1",
        "key": "MSCW",
        "values": [
          {
            "key": "W",
            "name": "Won't",
            "description": "Won't"
          },
          {
            "key": "C",
            "name": "Could",
            "description": "Could"
          },
          {
            "key": "S",
            "name": "Should",
            "description": "Should"
          },
          {
            "key": "M",
            "name": "Must",
            "description": "Must"
          }
        ]
      }
    }
  },
  "outcome": "x_basic:MSCW:1.0.0",
  "mapping": [
    {
      "ssvc:SINV:1.0.0": "FR",
      "ssvc:E:1.0.0": "N",
      "ssvc:PVA:1.0.0": "L",
      "x_basic:MSCW:1.0.0": "W"
    },
    {
      "ssvc:SINV:1.0.0": "FR",
      "ssvc:E:1.0.0": "N",
      "ssvc:PVA:1.0.0": "A",
      "x_basic:MSCW:1.0.0": "W"
    },
    {
      "ssvc:SINV:1.0.0": "FR",
      "ssvc:E:1.0.0": "N",
      "ssvc:PVA:1.0.0": "P",
      "x_basic:MSCW:1.0.0": "W"
    },
    {
      "ssvc:SINV:1.0.0": "FR",
      "ssvc:E:1.0.0": "P",
      "ssvc:PVA:1.0.0": "L",
      "x_basic:MSCW:1.0.0": "W"
    },
    {
      "ssvc:SINV:1.0.0": "FR",
      "ssvc:E:1.0.0": "P",
      "ssvc:PVA:1.0.0": "A",
      "x_basic:MSCW:1.0.0": "W"
    },
    {
      "ssvc:SINV:1.0.0": "FR",
      "ssvc:E:1.0.0": "P",
      "ssvc:PVA:1.0.0": "P",
      "x_basic:MSCW:1.0.0": "W"
    },
    {
      "ssvc:SINV:1.0.0": "FR",
      "ssvc:E:1.0.0": "A",
      "ssvc:PVA:1.0.0": "L",
      "x_basic:MSCW:1.0.0": "W"
    },
    {
      "ssvc:SINV:1.0.0": "FR",
      "ssvc:E:1.0.0": "A",
      "ssvc:PVA:1.0.0": "A",
      "x_basic:MSCW:1.0.0": "C"
    },
    {
      "ssvc:SINV:1.0.0": "FR",
      "ssvc:E:1.0.0": "A",
      "ssvc:PVA:1.0.0": "P",
      "x_basic:MSCW:1.0.0": "C"
    },
    {
      "ssvc:SINV:1.0.0": "C",
      "ssvc:E:1.0.0": "N",
      "ssvc:PVA:1.0.0": "L",
      "x_basic:MSCW:1.0.0": "C"
    },
    {
      "ssvc:SINV:1.0.0": "C",
      "ssvc:E:1.0.0": "N",
      "ssvc:PVA:1.0.0": "A",
      "x_basic:MSCW:1.0.0": "C"
    },
    {
      "ssvc:SINV:1.0.0": "C",
      "ssvc:E:1.0.0": "N",
      "ssvc:PVA:1.0.0": "P",
      "x_basic:MSCW:1.0.0": "C"
    },
    {
      "ssvc:SINV:1.0.0": "C",
      "ssvc:E:1.0.0": "P",
      "ssvc:PVA:1.0.0": "L",
      "x_basic:MSCW:1.0.0": "C"
    },
    {
      "ssvc:SINV:1.0.0": "C",
      "ssvc:E:1.0.0": "P",
      "ssvc:PVA:1.0.0": "A",
      "x_basic:MSCW:1.0.0": "C"
    },
    {
      "ssvc:SINV:1.0.0": "C",
      "ssvc:E:1.0.0": "P",
      "ssvc:PVA:1.0.0": "P",
      "x_basic:MSCW:1.0.0": "S"
    },
    {
      "ssvc:SINV:1.0.0": "C",
      "ssvc:E:1.0.0": "A",
      "ssvc:PVA:1.0.0": "L",
      "x_basic:MSCW:1.0.0": "S"
    },
    {
      "ssvc:SINV:1.0.0": "C",
      "ssvc:E:1.0.0": "A",
      "ssvc:PVA:1.0.0": "A",
      "x_basic:MSCW:1.0.0": "S"
    },
    {
      "ssvc:SINV:1.0.0": "C",
      "ssvc:E:1.0.0": "A",
      "ssvc:PVA:1.0.0": "P",
      "x_basic:MSCW:1.0.0": "S"
    },
    {
      "ssvc:SINV:1.0.0": "UU",
      "ssvc:E:1.0.0": "N",
      "ssvc:PVA:1.0.0": "L",
      "x_basic:MSCW:1.0.0": "S"
    },
    {
      "ssvc:SINV:1.0.0": "UU",
      "ssvc:E:1.0.0": "N",
      "ssvc:PVA:1.0.0": "A",
      "x_basic:MSCW:1.0.0": "S"
    },
    {
      "ssvc:SINV:1.0.0": "UU",
      "ssvc:E:1.0.0": "N",
      "ssvc:PVA:1.0.0": "P",
      "x_basic:MSCW:1.0.0": "S"
    },
    {
      "ssvc:SINV:1.0.0": "UU",
      "ssvc:E:1.0.0": "P",
      "ssvc:PVA:1.0.0": "L",
      "x_basic:MSCW:1.0.0": "M"
    },
    {
      "ssvc:SINV:1.0.0": "UU",
      "ssvc:E:1.0.0": "P",
      "ssvc:PVA:1.0.0": "A",
      "x_basic:MSCW:1.0.0": "M"
    },
    {
      "ssvc:SINV:1.0.0": "UU",
      "ssvc:E:1.0.0": "P",
      "ssvc:PVA:1.0.0": "P",
      "x_basic:MSCW:1.0.0": "M"
    },
    {
      "ssvc:SINV:1.0.0": "UU",
      "ssvc:E:1.0.0": "A",
      "ssvc:PVA:1.0.0": "L",
      "x_basic:MSCW:1.0.0": "M"
    },
    {
      "ssvc:SINV:1.0.0": "UU",
      "ssvc:E:1.0.0": "A",
      "ssvc:PVA:1.0.0": "A",
      "x_basic:MSCW:1.0.0": "M"
    },
    {
      "ssvc:SINV:1.0.0": "UU",
      "ssvc:E:1.0.0": "A",
      "ssvc:PVA:1.0.0": "P",
      "x_basic:MSCW:1.0.0": "M"
    }
  ]
}

```


</details>

<details>

<summary>JSON DecisionTable with obfuscation</summary>

## With Obfuscation

```python
dt = DecisionTable(...)
dt.obfuscate().model_dump_json(indent=2)
```
results in

```json
{
  "name": "Test Table",
  "description": "A test decision table",
  "namespace": "x_test",
  "version": "0.0.0",
  "schemaVersion": "1-0-1",
  "decision_point_group": {
    "version": "1.0.0",
    "schemaVersion": "1-0-1",
    "name": "Coordinator Publication",
    "description": "The decision points used by the coordinator during publication.",
    "decision_points": {
      "E8BA-19C5": {
        "name": "Supplier Involvement",
        "description": "What is the state of the supplier’s work on addressing the vulnerability?",
        "namespace": "ssvc",
        "version": "1.0.0",
        "schemaVersion": "1-0-1",
        "key": "SINV",
        "values": [
          {
            "key": "FR",
            "name": "Fix Ready",
            "description": "The supplier has provided a patch or fix."
          },
          {
            "key": "C",
            "name": "Cooperative",
            "description": "The supplier is actively generating a patch or fix; they may or may not have provided a mitigation or work-around in the mean time."
          },
          {
            "key": "UU",
            "name": "Uncooperative/Unresponsive",
            "description": "The supplier has not responded, declined to generate a remediation, or no longer exists."
          }
        ]
      },
      "1B67-1CDB": {
        "name": "Exploitation",
        "description": "The present state of exploitation of the vulnerability.",
        "namespace": "ssvc",
        "version": "1.0.0",
        "schemaVersion": "1-0-1",
        "key": "E",
        "values": [
          {
            "key": "N",
            "name": "None",
            "description": "There is no evidence of active exploitation and no public proof of concept (PoC) of how to exploit the vulnerability."
          },
          {
            "key": "P",
            "name": "PoC",
            "description": "One of the following cases is true: (1) private evidence of exploitation is attested but not shared; (2) widespread hearsay attests to exploitation; (3) typical public PoC in places such as Metasploit or ExploitDB; or (4) the vulnerability has a well-known method of exploitation."
          },
          {
            "key": "A",
            "name": "Active",
            "description": "Shared, observable, reliable evidence that the exploit is being used in the wild by real attackers; there is credible public reporting."
          }
        ]
      },
      "69CD-D12A": {
        "name": "Public Value Added",
        "description": "How much value would a publication from the coordinator benefit the broader community?",
        "namespace": "ssvc",
        "version": "1.0.0",
        "schemaVersion": "1-0-1",
        "key": "PVA",
        "values": [
          {
            "key": "L",
            "name": "Limited",
            "description": "Minimal value added to the existing public information because existing information is already high quality and in multiple outlets."
          },
          {
            "key": "A",
            "name": "Ampliative",
            "description": "Amplifies and/or augments the existing public information about the vulnerability, for example, adds additional detail, addresses or corrects errors in other public information, draws further attention to the vulnerability, etc."
          },
          {
            "key": "P",
            "name": "Precedence",
            "description": "The publication would be the first publicly available, or be coincident with the first publicly available."
          }
        ]
      },
      "34DB-9388": {
        "name": "MoSCoW",
        "description": "The MoSCoW (Must, Should, Could, Won't) outcome group.",
        "namespace": "x_basic",
        "version": "1.0.0",
        "schemaVersion": "1-0-1",
        "key": "MSCW",
        "values": [
          {
            "key": "W",
            "name": "Won't",
            "description": "Won't"
          },
          {
            "key": "C",
            "name": "Could",
            "description": "Could"
          },
          {
            "key": "S",
            "name": "Should",
            "description": "Should"
          },
          {
            "key": "M",
            "name": "Must",
            "description": "Must"
          }
        ]
      }
    }
  },
  "outcome": "34DB-9388",
  "mapping": [
    {
      "E8BA-19C5": "FR",
      "1B67-1CDB": "N",
      "69CD-D12A": "L",
      "34DB-9388": "W"
    },
    {
      "E8BA-19C5": "FR",
      "1B67-1CDB": "N",
      "69CD-D12A": "A",
      "34DB-9388": "W"
    },
    {
      "E8BA-19C5": "FR",
      "1B67-1CDB": "N",
      "69CD-D12A": "P",
      "34DB-9388": "W"
    },
    {
      "E8BA-19C5": "FR",
      "1B67-1CDB": "P",
      "69CD-D12A": "L",
      "34DB-9388": "W"
    },
    {
      "E8BA-19C5": "FR",
      "1B67-1CDB": "P",
      "69CD-D12A": "A",
      "34DB-9388": "W"
    },
    {
      "E8BA-19C5": "FR",
      "1B67-1CDB": "P",
      "69CD-D12A": "P",
      "34DB-9388": "W"
    },
    {
      "E8BA-19C5": "FR",
      "1B67-1CDB": "A",
      "69CD-D12A": "L",
      "34DB-9388": "W"
    },
    {
      "E8BA-19C5": "FR",
      "1B67-1CDB": "A",
      "69CD-D12A": "A",
      "34DB-9388": "C"
    },
    {
      "E8BA-19C5": "FR",
      "1B67-1CDB": "A",
      "69CD-D12A": "P",
      "34DB-9388": "C"
    },
    {
      "E8BA-19C5": "C",
      "1B67-1CDB": "N",
      "69CD-D12A": "L",
      "34DB-9388": "C"
    },
    {
      "E8BA-19C5": "C",
      "1B67-1CDB": "N",
      "69CD-D12A": "A",
      "34DB-9388": "C"
    },
    {
      "E8BA-19C5": "C",
      "1B67-1CDB": "N",
      "69CD-D12A": "P",
      "34DB-9388": "C"
    },
    {
      "E8BA-19C5": "C",
      "1B67-1CDB": "P",
      "69CD-D12A": "L",
      "34DB-9388": "C"
    },
    {
      "E8BA-19C5": "C",
      "1B67-1CDB": "P",
      "69CD-D12A": "A",
      "34DB-9388": "C"
    },
    {
      "E8BA-19C5": "C",
      "1B67-1CDB": "P",
      "69CD-D12A": "P",
      "34DB-9388": "S"
    },
    {
      "E8BA-19C5": "C",
      "1B67-1CDB": "A",
      "69CD-D12A": "L",
      "34DB-9388": "S"
    },
    {
      "E8BA-19C5": "C",
      "1B67-1CDB": "A",
      "69CD-D12A": "A",
      "34DB-9388": "S"
    },
    {
      "E8BA-19C5": "C",
      "1B67-1CDB": "A",
      "69CD-D12A": "P",
      "34DB-9388": "S"
    },
    {
      "E8BA-19C5": "UU",
      "1B67-1CDB": "N",
      "69CD-D12A": "L",
      "34DB-9388": "S"
    },
    {
      "E8BA-19C5": "UU",
      "1B67-1CDB": "N",
      "69CD-D12A": "A",
      "34DB-9388": "S"
    },
    {
      "E8BA-19C5": "UU",
      "1B67-1CDB": "N",
      "69CD-D12A": "P",
      "34DB-9388": "S"
    },
    {
      "E8BA-19C5": "UU",
      "1B67-1CDB": "P",
      "69CD-D12A": "L",
      "34DB-9388": "M"
    },
    {
      "E8BA-19C5": "UU",
      "1B67-1CDB": "P",
      "69CD-D12A": "A",
      "34DB-9388": "M"
    },
    {
      "E8BA-19C5": "UU",
      "1B67-1CDB": "P",
      "69CD-D12A": "P",
      "34DB-9388": "M"
    },
    {
      "E8BA-19C5": "UU",
      "1B67-1CDB": "A",
      "69CD-D12A": "L",
      "34DB-9388": "M"
    },
    {
      "E8BA-19C5": "UU",
      "1B67-1CDB": "A",
      "69CD-D12A": "A",
      "34DB-9388": "M"
    },
    {
      "E8BA-19C5": "UU",
      "1B67-1CDB": "A",
      "69CD-D12A": "P",
      "34DB-9388": "M"
    }
  ]
}

```

</details>

data/schema/v2/Decision_Table-2-0-0.schema.json
```diff
@@ -0,0 +1,107 @@
+{
+  "title": "SsvcDecisionPoint",
```
sei-vsarvepalli: This seems to the wrong file? It is not the Decision Table Schema -

src/ssvc/__init__.py
```diff
@@ -19,5 +19,16 @@
...
+# import all submodules from ssvc.decision_points and ssvc.outcomes to populate the registries
+# automatically walk through the decision_points and outcomes directories
+# dive into each submodule and import all its parts
+import_modules(IMPORTABLES, include_children=True)
+
+if __name__ == "__main__":
+    # confirms that the registries are populated
+    from ssvc.registry import REGISTRY
+
+    print(REGISTRY.model_dump_json(indent=2))
```
sei-vsarvepalli, line 34:
```suggestion
#    print(REGISTRY.model_dump_json(indent=2))
```

Probably should comment this out?

sei-vsarvepalli:
on import of SSVC look like the mapped tables are being printed.
```
>>> import ssvc
{'ssvc:SI:1.0.0': 'N', 'ssvc:MI:2.0.0': 'D'}
{'ssvc:SI:1.0.0': 'N', 'ssvc:MI:2.0.0': 'MSC'}
{'ssvc:SI:1.0.0': 'N', 'ssvc:MI:2.0.0': 'MEF'}
{'ssvc:SI:1.0.0': 'N', 'ssvc:MI:2.0.0': 'MF'}
{'ssvc:SI:1.0.0': 'M', 'ssvc:MI:2.0.0': 'D'}
{'ssvc:SI:1.0.0': 'M', 'ssvc:MI:2.0.0': 'MSC'}
{'ssvc:SI:1.0.0': 'M', 'ssvc:MI:2.0.0': 'MEF'}
{'ssvc:SI:1.0.0': 'M', 'ssvc:MI:2.0.0': 'MF'}
{'ssvc:SI:1.0.0': 'J', 'ssvc:MI:2.0.0': 'D'}
{'ssvc:SI:1.0.0': 'J', 'ssvc:MI:2.0.0': 'MSC'}
{'ssvc:SI:1.0.0': 'J', 'ssvc:MI:2.0.0': 'MEF'}
{'ssvc:SI:1.0.0': 'J', 'ssvc:MI:2.0.0': 'MF'}
{'ssvc:SI:1.0.0': 'H', 'ssvc:MI:2.0.0': 'D'}
{'ssvc:SI:1.0.0': 'H', 'ssvc:MI:2.0.0': 'MSC'}
{'ssvc:SI:1.0.0': 'H', 'ssvc:MI:2.0.0': 'MEF'}
{'ssvc:SI:1.0.0': 'H', 'ssvc:MI:2.0.0': 'MF'}
{'ssvc:SI:1.0.0': 'C', 'ssvc:MI:2.0.0': 'D'}
{'ssvc:SI:1.0.0': 'C', 'ssvc:MI:2.0.0': 'MSC'}
{'ssvc:SI:1.0.0': 'C', 'ssvc:MI:2.0.0': 'MEF'}
{'ssvc:SI:1.0.0': 'C', 'ssvc:MI:2.0.0': 'MF'}
```

data/schema/v2/Decision_Table-2-0-0.schema.json
sei-vsarvepalli: It will be good to fix this - although the automation might generate it too.

data/schema/v2/Ssvc_Object_Registry-2-0-0.schema.json
```diff
@@ -0,0 +1,261 @@
...
+      "description": "A dictionary mapping type names to NsType objects.",
+      "type": "object"
+    }
+  },
+  "required": [
+    "name",
+    "description",
+    "schemaVersion"
+  ]
+}
```
sei-vsarvepalli, line 261:
```suggestion
{
  "$defs": {
    "DecisionPoint": {
      "description": "Models a single decision point as a list of values.\n\nDecision points should have the following attributes:\n\n- name (str): The name of the decision point\n- description (str): A description of the decision point\n- version (str): A semantic version string for the decision point\n- namespace (str): The namespace (a short, unique string): For example, \"ssvc\" or \"cvss\" to indicate the source of the decision point\n- key (str): A key (a short, unique string within the namespace) that can be used to identify the decision point in a shorthand way\n- values (tuple): A tuple of DecisionPointValue objects",
      "properties": {
        "namespace": {
          "description": "The namespace of the SSVC object.",
          "examples": [
            "ssvc",
            "cisa",
            "x_com.example//com.example#private",
            "ssvc/de-DE/example.organization#reference-arch-1"
          ],
          "maxLength": 1000,
          "minLength": 3,
          "pattern": "^(?=.{3,1000}$)(?:x_(?!.*[.-]{2,})[a-z][a-z0-9]+(?:[.-][a-z0-9]+)*|(?!.*[.-]{2,})[a-z][a-z0-9]+(?:[.-][a-z0-9]+)*)(?:(?:\\/(([A-Za-z]{2,3}(-[A-Za-z]{3}(-[A-Za-z]{3}){0,2})?|[A-Za-z]{4,8})(-[A-Za-z]{4})?(-([A-Za-z]{2}|[0-9]{3}))?(-([A-Za-z0-9]{5,8}|[0-9][A-Za-z0-9]{3}))*(-[A-WY-Za-wy-z0-9](-[A-Za-z0-9]{2,8})+)*(-[Xx](-[A-Za-z0-9]{1,8})+)?|[Xx](-[A-Za-z0-9]{1,8})+|[Ii]-[Dd][Ee][Ff][Aa][Uu][Ll][Tt]|[Ii]-[Mm][Ii][Nn][Gg][Oo])\\/|\\/\\/)(?!.*[.-]{2,})[a-zA-Z][a-zA-Z0-9]*(?:[.-][a-zA-Z0-9]+)*(?:#[a-zA-Z0-9]+(?:[.-][a-zA-Z0-9]+)*)?(?:\\/(?!.*[.-]{2,})[a-zA-Z][a-zA-Z0-9]*(?:[.-][a-zA-Z0-9]+)*(?:#[a-zA-Z0-9]+(?:[.-][a-zA-Z0-9]+)*)?)*)?$",
          "title": "Namespace",
          "type": "string"
        },
        "key": {
          "description": "A short, non-empty string identifier for the object. Keys must start with an alphanumeric, contain only alphanumerics and `_`, and end with an alphanumeric.(`T*` is explicitly grandfathered in as a valid key, but should not be used for new objects.)",
          "examples": [
            "E",
            "A",
            "SI",
            "L",
            "M",
            "H",
            "Mixed_case_OK",
            "alph4num3ric"
          ],
          "minLength": 1,
          "pattern": "^(([a-zA-Z0-9])|([a-zA-Z0-9][a-zA-Z0-9_]*[a-zA-Z0-9])|(T\\*))$",
          "title": "Key",
          "type": "string"
        },
        "version": {
          "default": "0.0.1",
          "description": "The version of the SSVC object. This must be a valid semantic version string.",
          "examples": [
            "1.0.0",
            "2.1.3"
          ],
          "minLength": 5,
          "pattern": "^(0|[1-9]\\d*)\\.(0|[1-9]\\d*)\\.(0|[1-9]\\d*)(?:-((?:0|[1-9]\\d*|\\d*[a-zA-Z-][0-9a-zA-Z-]*)(?:\\.(?:0|[1-9]\\d*|\\d*[a-zA-Z-][0-9a-zA-Z-]*))*))?(?:\\+([0-9a-zA-Z-]+(?:\\.[0-9a-zA-Z-]+)*))?$",
          "title": "Version",
          "type": "string"
        },
        "name": {
          "title": "Name",
          "type": "string"
        },
        "description": {
          "title": "Description",
          "type": "string"
        },
        "schemaVersion": {
          "const": "2.0.0",
          "title": "Schemaversion",
          "type": "string"
        },
        "values": {
          "items": {
            "$ref": "#/$defs/DecisionPointValue"
          },
          "title": "Values",
          "type": "array"
        }
      },
      "required": [
        "namespace",
        "key",
        "name",
        "description",
        "schemaVersion",
        "values"
      ],
      "title": "DecisionPoint",
      "type": "object"
    },
    "DecisionPointValue": {
      "description": "Models a single value option for a decision point.\n\nEach value should have the following attributes:\n\n- name (str): A name\n- description (str): A description\n- key (str): A key (a short, unique string) that can be used to identify the value in a shorthand way\n- _comment (str): An optional comment that will be included in the object.",
      "properties": {
        "key": {
          "description": "A short, non-empty string identifier for the object. Keys must start with an alphanumeric, contain only alphanumerics and `_`, and end with an alphanumeric.(`T*` is explicitly grandfathered in as a valid key, but should not be used for new objects.)",
          "examples": [
            "E",
            "A",
            "SI",
            "L",
            "M",
            "H",
            "Mixed_case_OK",
            "alph4num3ric"
          ],
          "minLength": 1,
          "pattern": "^(([a-zA-Z0-9])|([a-zA-Z0-9][a-zA-Z0-9_]*[a-zA-Z0-9])|(T\\*))$",
          "title": "Key",
          "type": "string"
        },
        "name": {
          "title": "Name",
          "type": "string"
        },
        "description": {
          "title": "Description",
          "type": "string"
        }
      },
      "required": [
        "key",
        "name",
        "description"
      ],
      "title": "DecisionPointValue",
      "type": "object"
    }
  },
  "description": "DecisionTable: A flexible, serializable SSVC decision table model.\n\nThis model represents a decision table that can be used to map combinations of decision point values\nto outcomes. It allows for flexible mapping and can be used with helper methods to generate DataFrame and CSV representations\nof the decision table.\n\nAttributes:",
  "properties": {
    "name": {
      "title": "Name",
      "type": "string"
    },
    "description": {
      "title": "Description",
      "type": "string"
    },
    "namespace": {
      "description": "The namespace of the SSVC object.",
      "examples": [
        "ssvc",
        "cisa",
        "x_com.example//com.example#private",
        "ssvc/de-DE/example.organization#reference-arch-1"
      ],
      "maxLength": 1000,
      "minLength": 3,
      "pattern": "^(?=.{3,1000}$)(?:x_(?!.*[.-]{2,})[a-z][a-z0-9]+(?:[.-][a-z0-9]+)*|(?!.*[.-]{2,})[a-z][a-z0-9]+(?:[.-][a-z0-9]+)*)(?:(?:\\/(([A-Za-z]{2,3}(-[A-Za-z]{3}(-[A-Za-z]{3}){0,2})?|[A-Za-z]{4,8})(-[A-Za-z]{4})?(-([A-Za-z]{2}|[0-9]{3}))?(-([A-Za-z0-9]{5,8}|[0-9][A-Za-z0-9]{3}))*(-[A-WY-Za-wy-z0-9](-[A-Za-z0-9]{2,8})+)*(-[Xx](-[A-Za-z0-9]{1,8})+)?|[Xx](-[A-Za-z0-9]{1,8})+|[Ii]-[Dd][Ee][Ff][Aa][Uu][Ll][Tt]|[Ii]-[Mm][Ii][Nn][Gg][Oo])\\/|\\/\\/)(?!.*[.-]{2,})[a-zA-Z][a-zA-Z0-9]*(?:[.-][a-zA-Z0-9]+)*(?:#[a-zA-Z0-9]+(?:[.-][a-zA-Z0-9]+)*)?(?:\\/(?!.*[.-]{2,})[a-zA-Z][a-zA-Z0-9]*(?:[.-][a-zA-Z0-9]+)*(?:#[a-zA-Z0-9]+(?:[.-][a-zA-Z0-9]+)*)?)*)?$",
      "title": "Namespace",
      "type": "string"
    },
    "version": {
      "default": "0.0.1",
      "description": "The version of the SSVC object. This must be a valid semantic version string.",
      "examples": [
        "1.0.0",
        "2.1.3"
      ],
      "minLength": 5,
      "pattern": "^(0|[1-9]\\d*)\\.(0|[1-9]\\d*)\\.(0|[1-9]\\d*)(?:-((?:0|[1-9]\\d*|\\d*[a-zA-Z-][0-9a-zA-Z-]*)(?:\\.(?:0|[1-9]\\d*|\\d*[a-zA-Z-][0-9a-zA-Z-]*))*))?(?:\\+([0-9a-zA-Z-]+(?:\\.[0-9a-zA-Z-]+)*))?$",
      "title": "Version",
      "type": "string"
    },
    "schemaVersion": {
      "const": "2.0.0",
      "title": "Schemaversion",
      "type": "string"
    },
    "decision_points": {
      "additionalProperties": {
        "$ref": "#/$defs/DecisionPoint"
      },
      "description": "A non-empty dictionary of decision points Decision point IDs are recommended as keys.",
      "minProperties": 1,
      "title": "Decision Points",
      "type": "object"
    },
    "outcome": {
      "description": "The key of the decision point in `self.decision_points` that represents the outcome of the decision table.",
      "minLength": 1,
      "title": "Outcome",
      "type": "string"
    },
    "mapping": {
      "description": "Mapping of decision point values to outcomes.",
      "items": {
        "additionalProperties": {
          "type": "string"
        },
        "type": "object"
      },
      "title": "Mapping",
      "type": "array"
    }
  },
  "required": [
    "name",
    "description",
    "namespace",
    "schemaVersion",
    "decision_points",
    "outcome"
  ],
  "title": "DecisionTable",
  "type": "object"
}
```

src/ssvc/decision_tables/ssvc/human_impact.py
```diff
@@ -0,0 +1,165 @@
...
+from ssvc.decision_points.ssvc.human_impact import LATEST as HumanImpact
+from ssvc.decision_points.ssvc.mission_impact import LATEST as MissionImpact
+from ssvc.decision_points.ssvc.safety_impact import SAFETY_IMPACT_1 as SituatedSafetyImpact
+from ssvc.decision_tables.base import DecisionTable, dpdict_to_combination_list
+from ssvc.decision_tables.helpers import write_csv
+
+dp_dict = {dp.id: dp for dp in [SituatedSafetyImpact, MissionImpact, HumanImpact]}
+
+for combo in dpdict_to_combination_list(dp_dict,exclude=HumanImpact.id):
+    print(combo)
```
sei-vsarvepalli:
```suggestion
#for combo in dpdict_to_combination_list(dp_dict,exclude=HumanImpact.id):
#    print(combo)
```

Avoid to reduce noise or move this to a def for debugging

src/ssvc/decision_tables/helpers.py
```diff
@@ -0,0 +1,60 @@
...
+#  Licensed under a MIT (SEI)-style license, please see LICENSE or contact
+#  permission@sei.cmu.edu for full terms.
+#  [DISTRIBUTION STATEMENT A] This material has been approved for
+#  public release and unlimited distribution. Please see Copyright notice
+#  for non-US Government use and distribution.
+#  This Software includes and/or makes use of Third-Party Software each
+#  subject to its own license.
+#  DM24-0278
+
+from ssvc.decision_tables.base import decision_table_to_longform_df
```
sei-vsarvepalli, line 25: Whe I try to run this in the relative path of the repository it fails as there is also another `ssvc` directory under here in "./" of this path.  Basically the `src/ssvc/decision_tables/ssvc` is in the preferred order for python file execution over the `src/ssvc` path.

```
(venv) Mac-redjayapple:SSVC vssarvepalli$ python src/ssvc/decision_tables/base.py 
Traceback (most recent call last):
  File "/Users/vijay/Development/SSVC/src/ssvc/decision_tables/base.py", line 31, in <module>
    from ssvc._mixins import _Base, _Commented, _Namespaced, _SchemaVersioned, _Versioned
ModuleNotFoundError: No module named 'ssvc._mixins'
(venv) Mac-redjayapple:SSVC vssarvepalli$ 
```

sei-vsarvepalli:
We should add a helper method perhaps in decision_point/helper.py that helps get a Decision Point Value using "Name" or "Key" . Something like below or we could use as iteration is needed from current Decision Table model.

```
def dpvalueLookup(dp: SsvcDecisionPoint, name: str = None, key: str = None) -> DecisionPointValue:
""" Find Decision Point Values by key or by name """
    for dv in dp.values:
        if name and dv.name.lower() == name.lower():
            return dv
        if key and dv.key == key:
            return dv
    return None

```

data/schema/v2/Decision_Table-2-0-0.schema.json
ahouseholder: Yeah. It works better when we use the right object name 😁

src/ssvc/decision_tables/ssvc/human_impact.py
ahouseholder: yeah it was a debugging thing. Gone now.

src/ssvc/registry/base.py
```diff
@@ -0,0 +1,424 @@
...
+
+
+class NonValuedVersion(BaseModel):
+    version: VersionString
+    obj: _GenericSsvcObject
+
+
+class ValuedVersion(BaseModel):
+    version: VersionString
+    obj: _GenericSsvcObject
```
ahouseholder: The `obj: _GenericSsvcObject` lines are a problem because it's not copying the full object, just the fields that are explicitly in the `_GenericSsvcObject` mixin. We need to resolve this before merging.

sei-vsarvepalli:
Also CVSS Equivalency Set to CVSS LMHC scoring also fails the pydantic order validation test. The sorted array is enclosed here for testing. 

[cvss_v4_mapping.json](https://github.com/user-attachments/files/21554128/cvss_v4_mapping.json)

sei-vsarvepalli (approved):

Issue #818, opened 2025-07-11
PDF Exports cut off at 1 page


0x4rk0:
**Describe the bug**
When exporting an SSVC decision, and inputting a CVE, if the CVE description is long, the PDF output is cutoff to 1 page. Example CVE: CVE-2025-3648

**To Reproduce**

1. Fill out SSVC Decision Tree
2. Input a CVE at the export stage
3. If the CVE details are long enough, the SSVC details are cut off to 1 page. Examples CVE to replicate is CVE-2025-3648.

**Expected behavior**
Multiple pages that contain all information pertaining to an SSVC decision. Example would be to export a blank SSVC with no CVE.

**Screenshots**

<img width="1512" height="819" alt="Image" src="https://github.com/user-attachments/assets/95843284-b381-40cc-b525-fa4bdc20adcf" />

**Platform details**
This has occurred on multiple browsers and operating systems.

**Additional context**
N/A

0x4rk0:
Hi there,

Sorry for the delay, this was using https://www.cisa.gov/ssvc-calculator.

PR #822, opened 2025-07-15
Fix for Bug paging issue #818

sei-vsarvepalli:
This is to fix issue #818 paging issue with PDF exporter.


sei-renae (approved):
Used the steps to reproduce #818 to successfully create a multi-page PDF using a branch for #822

PR #841, opened 2025-08-05
Update CISA Track* to Monitor and version

### LINKED ISSUE #699: Change "Track *" to "Monitor" in our representation of CISA's tree

CISA's adoption of SSVC resulted in their own customized outcome set that they use internally. However, some of the terminology used is highly idiosyncratic to their particular environment as a [CSIRT with National Responsibility](https://insights.sei.cmu.edu/library/natcsirt-resources/). In particular, terms like "Track *" have no significant meaning outside of their environment.

We include in our code and documentation examples derived from CISA's SSVC implementation. However, it's not necessary for us to retain their exact terminology if we can generalize concepts to be more applicable to other users of SSVC.

The suggestion here is to replace "Track *" with something more generally understood, like "Monitor". 

**Additional context**

This change would require us to modify at least:

- the json representation of the CISA model
- the CISA outcome group in the python code (which will flow down to the examples in https://certcc.github.io/SSVC/reference/code/outcomes/?h=cisa )

sei-vsarvepalli:
- resolves #699

CISA uses 2.0.3 for the Decision Table Version as the current one in various places. The next version 2.0.4 is expected to get rid of Track* in favor of the Monitor as guidance provided by Jono in May 2025.

### DIFF
--- a/src/ssvc/outcomes/cisa/scoring.py
+++ b/src/ssvc/outcomes/cisa/scoring.py
@@ -33,6 +33,14 @@ _TRACK_STAR = DecisionPointValue(
     "CISA recommends remediating Track* vulnerabilities within standard update timelines.",
 )
 
+
+_MONITOR = DecisionPointValue(
+    name="Monitor",
+    key="M",
+    description="The vulnerability contains specific characteristics that may require closer monitoring for changes. "
+    "CISA recommends remediating Monitor vulnerabilities within standard update timelines.",
+)
+
 _ATTEND = DecisionPointValue(
     name="Attend",
     key="A",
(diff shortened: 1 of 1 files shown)


ahouseholder:
Converting to draft to avoid premature merge.

sei-vsarvepalli:
> What's different between this and
> 
> * [Replace `Track*`/`Track *` with `Monitor` in CISA-based decision model (2nd try) #742](https://github.com/CERTCC/SSVC/pull/742)
> 
> ?
> 
> I had been keeping #742 in sync as we merged into main, but holding it as a _draft PR_ so we didn't accidentally merge it until CISA had their changes ready. Also, #742 has more changes in it than just the outcome object in python.

Yeh the #742 diverged too much away from this base. So I just moved to make a smaller PR specific only to this issue - programmatically. It is required for the Decision Tables that I am generating to work properly.

sei-vsarvepalli:
Hello @j--- 

Can we get your help in advancing resolution to this? Our hope for September publish release is that this is consistent with CISA and is accurate, especially in the public forums.

sei-vsarvepalli:
I think this issue is moot now with BOD-26-04 taking precedence.

PR #882, opened 2025-08-15
implement new namespace patterns based on ABNF

bernhardreiter:
* add the new leading ABNF from #858 for namespace into `src/ssvc/utils/ssvc_namespace_pattern.abnf`
* replace old regular expression patterns in `patterns.py` with ones generated from the ABNF
* replace JSON schema files with the newly generated ones
* adapt tests for the changed namespace rules

based on #881, so that tests run in principle.

On failing test was there before and is untouched. 

A review should look at the changed testcases in test_namespaces_pattern.py
and note that [docs/reference/code/namespaces.md](https://github.com/CERTCC/SSVC/blob/main/docs/reference/code/namespaces.md) is know to need to be overhauled with the results of the discussions that lead to #858.

Assuming shared copyright on the new file, `ssvc_namespace_pattern.abnf`. As I am working on behalf of @tschmidtb51, the (German) BSI owns the copyright of copyrightable changes. 

resolve #858

PR #898, opened 2025-08-20
Namespacepattern updates from #882

sei-vsarvepalli:
This is a replica #882 to get rid of test failures.


ahouseholder (approved):
This all looks good. The only change I made was to have our unit tests use the newly-official `test` namespace instead of `x_example.test#test`.

PR #915, opened 2025-08-26
Add 'invalid' to the registered namespaces

### LINKED ISSUE #907: Reserve registered namespace `invalid`
While implementing, I realized that we come up with any (in)valid decision point inside the registered namespace `test`. However, we don't have a place to test namespace things with a registered namespace that is invalid - i.e. not registered.

I think it would be helpful to reserve the registered namespace `invalid` - comparable to the `.invalid` assignment in DNS.

**Describe alternatives you've considered**
Using random namespace that are not yet registered: You can't be sure that they won't be registered in the future.

**Additional context**
related to #802 (and https://github.com/oasis-tcs/csaf/issues/1152 (Test 6.1.48))

ahouseholder:
- resolves #907 

The original issue also mentions `x_invalid`, however, if we add `x_invalid` to the registered namespaces, it will fail to match the namespace pattern because the pattern assumes anything that starts with `x_` will have a fragment. So I'm leaving that out of this PR in the interest of keeping the PR simple.

### DIFF
--- a/src/ssvc/namespaces.py
+++ b/src/ssvc/namespaces.py
@@ -61,7 +61,6 @@ class NameSpace(StrEnum):
     EXAMPLE = auto()
     TEST = auto()
     INVALID = auto()
-    X_INVALID = auto()
 
     @classmethod
     def validate(cls, value: str) -> str:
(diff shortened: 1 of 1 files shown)


tschmidtb51 (changes requested):
Please see: https://github.com/CERTCC/SSVC/issues/907#issuecomment-3225857383

ahouseholder:
Withdrawing this PR until we get to resolution on #907

PR #917, opened 2025-08-27
Remove default volume mount in docker-compose config

ahouseholder:
Remove a default `volume` mount configuration from `docker-compose.yml` because it was causing bad interactions between the host system and the container when `uv` would rebuild the environment within the container. The MacOS dev host was flagging python libraries as untrusted after running the containers.

(cherry picked from commit 51f345e9840ccd0eafb44e68b4a1032fa7ddc4d9)

## Copilot Summary

This pull request makes small changes to the `docker/docker-compose.yml` file by removing the bind mount volumes from the `docs` and `registry_api` services. This means the application source code will no longer be mounted into these containers at runtime, likely to improve container isolation or deployment consistency.

### DIFF
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -32,8 +32,6 @@ services:
     image: docs:latest
     depends_on:
       - dependencies
-    volumes:
-      - ..:/app
     ports:
       - "8000:8000"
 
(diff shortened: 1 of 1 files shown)


ahouseholder:
bypassing approval and merging because this one is a small change and is basically blocking me (and _only_ me)

PR #967, opened 2025-09-15
Use pattern string instead of compiled pattern in field_specs.py

### LINKED ISSUE #965: Use pattern string instead of compiled pattern in `field_spec.py`
Based on this comment:

> Even better would be to have pydantic's conversion code checking these and flagging problems.
> 
> As a follow-up I've created https://github.com/pydantic/pydantic/issues/12249 to improve the upstream situation.
> 
> I've noticed that the situation would have been better, if https://github.com/CERTCC/SSVC/blob/7fd1697c7b34f991754a716ef665fe1396a7bd59/src/ssvc/utils/field_specs.py#L53 would not have used the compiled pattern, but the pattern string. Turns out that pydantic treats a compiled pattern differently than a pattern string.
> 
>  

 _Originally posted by @bernhardreiter in [#916](https://github.com/CERTCC/SSVC/issues/916#issuecomment-3293532294)_

The task here is to replace our use of the compiled `NS_PATTERN` with the `NS_PATTERN.pattern` string.

ahouseholder:
- resolves #965 

This PR replaces the compiled NS_PATTERN with the pattern string when creating the field specs.

A side effect of this change is that the pattern no longer needs the lookahead for length check, as we're specifying as part of the pydantic class and that carries into the JSON schema directly without it having to be part of the regex.

## Copilot Summary

This pull request simplifies and standardizes the namespace pattern validation logic across several schema and utility files. The main change is the removal of an unnecessary length check from the regular expression pattern, relying instead on explicit `minLength` and `maxLength` constraints. This makes the regex easier to maintain and ensures consistency between schema validation and Python-side checks.

Most important changes:

**Schema pattern updates:**

* Removed the inline length check (`(?=.{3,1000}$)`) from the `pattern` field in the JSON schema definitions for `DecisionPoint_2_0_0.schema.json`, `DecisionPointGroup_2_0_0.schema.json`, `DecisionTable_2_0_0.schema.json`, `SelectionList_2_0_0.schema.json`, and `SsvcObjectRegistry_2_0_0.schema.json`, so that the regex only enforces the namespace format, not its length. Length is now enforced solely by the `minLength` and `maxLength` schema properties. [[1]](diffhunk://#diff-7e98917be3ea68a0a9040eaf8289c7a8a38056718ef75f0098685a613baf8e2dL58-R58) [[2]](diffhunk://#diff-e903fbbf1a71c864795daa928b1dd17881b82c56b25f5a633f2974e74136c22dL23-R23) [[3]](diffhunk://#diff-e571c74d463127d0ede709adf8e98575dd38b5117b100b7e84ab04f731a64207L23-R23) [[4]](diffhunk://#diff-e571c74d463127d0ede709adf8e98575dd38b5117b100b7e84ab04f731a64207L136-R136) [[5]](diffhunk://#diff-7582a0c78b81d7156ca61f4f9d41b342fd38089f53463e7ccb446819160cc0e8L82-R82) [[6]](diffhunk://#diff-dfe09f144c0c96260942b0d559b1ba8110f48ccdbcaf9f44258243fc8153aa01L23-R23) [[7]](diffhunk://#diff-dfe09f144c0c96260942b0d559b1ba8110f48ccdbcaf9f44258243fc8153aa01L138-R138)

**Python utility and test consistency:**

* Updated the namespace regex in `src/ssvc/utils/patterns.py` to remove the length check from the pattern string, matching the schema change.
* Changed `src/ssvc/utils/field_specs.py` to use the raw pattern string from the compiled regex, ensuring consistency between the schema and Python validation.
* Improved the namespace pattern unit test in `src/test/test_namespaces.py` to accumulate and report all unexpected matches, making debugging easier.


sei-vsarvepalli (approved):
Well.. the tests are running okay. This regex is one I cannot read :)

bernhardreiter:
What about using NS_PATTERN_STR instead of NS_PATTERN.pattern and remove NS_PATTERN entirely? It would be an additional simplification.

> This regex is one I cannot read :)

If you look at the different parts of the regex in pattern.py, that is close to the ABNF and can be understood. The one in NS_PATTERN_STR is just unrolled. ;)

PR #969, opened 2025-09-15
Updates and small fixes, move ascii_tree to helpers

sei-vsarvepalli:
This does not completely move to the proposed HTML format, but has the helper method added dt2df_html that we can decide to use later. 

There are also other small fixes here
1. `docs/howto/supplier_tree.md` runaway sting fixed
2. `src/README.cd` move the ascii_tree method to helpers where it belongs instead of base.py
3. update to both `src/ssvc/decision_tables/base.py` and `src/ssvc/decision_tables/helpers.py` (type hinting on df2dt_md also corrected) 
```diff 
 def dt2df_md(
-    dt: "DecisionTable",
+    dt: DecisionTable,
     longform: bool = True,
 ) -> str:
```


ahouseholder (approved):
Looks good to me.

We should have a separate conversation about whether we want to move some of these functions into `DecisionTable` class methods, or if we want to try to keep data objects separate from functions. There are good reasons for both, but we should probably pick one as an architecture choice and be consistent with it.

PR #980, opened 2025-09-19
Move pyproject.toml to root level dir

### LINKED ISSUE #974: Move `/src/pyproject.toml` to `/pyproject.toml`
It would be easier to work with uv commands if we didn't have to remember to always specify `--project=src` when doing any `uv foo` things. This isn't as big a deal when we're doing things via `Dockerfile` or `Makefile` or the like, but it's bitten me a few times when using `uv` at the command line. I think it's probably pretty straightforward to move `./src/pyproject.toml` to `./pyproject.toml` then clean up any `uv foo --project=src` commands and be done with it once and for all. I might be wrong of course, but if we could do this it'd make life a smidge easier for future us.

sei-renae:
Resolves #974 

PR in draft state to test GH workflows


sei-vsarvepalli (approved):
sure thing.

pyproject.toml
```diff
@@ -62,13 +62,13 @@ ssvc_doctools="ssvc.doctools:main"
 
 [tool.setuptools.packages.find]
 where = ["."]  # list of folders that contain the packages (["."] by default)
-include = ["ssvc*"]  # package names should match these glob patterns (["*"] by default)
+include = ["src*"]  # package names should match these glob patterns (["*"] by default)
```
ahouseholder: I think we might want this to be `src/ssvc/*` ?

ahouseholder (changes requested):
I don't see any changes to `docker/Dockerfile` or `Makefile`. Those will break if left unmodified.

pyproject.toml
ahouseholder: or maybe it's `where = ["./src"]` and leave `include = ["ssvc*"]` ?

sei-renae: All 3 options successfully run `uv sync`

sei-renae:
> On this front, #986 makes the cut over to `uv` but since this is changing the location of `pyproject.toml` we'll still need to update it for that. (remove `- "/src"` from `directories:`)

How did #986 not break dependabot?

ahouseholder:
> How did #986 not break dependabot?

I think basically this line?
https://github.com/CERTCC/SSVC/pull/986/files#diff-dd4fbda47e51f1e35defb9275a9cd9c212ecde0b870cba89ddaaae65c5f3cd28R9


I added the `/src` to tell Dependabot where to look for pyproject.toml.

docker/Dockerfile
ahouseholder, line 1: I'm not sure about the changes in lines 12-16 here. It seems like it would end up copying the `/docker` directory to `/app` instead of `/docker/..`

PR #1002, opened 2025-10-07
Extend methods to generate Selection Schema for CVE and other projects

sei-vsarvepalli:
Added extension methods to support export of SSVC Selection data into schema supported format.

This pull request introduces enhancements to the serialization logic for selection models in `src/ssvc/selection.py`. The changes ensure that dumped data is consistently formatted, with empty or invalid values removed and timestamps properly formatted in JSON outputs.

Key improvements to serialization and data cleanup:

* Added a `_post_process` method to ensure all `Selection.values` are lists, remove empty or falsy items, and clean up empty array fields from the top-level dictionary.
* Overrode the `model_dump` and `model_dump_json` methods to apply the post-processing step before returning data, and to format timestamps as UTC RFC3339 strings in JSON output.

### DIFF
--- a/src/ssvc/selection.py
+++ b/src/ssvc/selection.py
@@ -312,7 +312,51 @@ class SelectionList(_SchemaVersioned, _Timestamped, BaseModel):
         schema = strip_nullable_anyof(schema)
 
         return order_schema(schema)
+    def _post_process(self, data):
+        """
+        Ensures all Selection.values are lists and removes empty array elements.
+        """
+        def fix_selection(selection):
+            # Convert tuple to list and filter out empty items
+            values = selection.get("values", [])
+            # Ensure it's a list, filter out empty/falsy items
+            selection["values"] = [v for v in list(values) if v]
+            return selection
 
+        # If this is a dict with selections, process each selection
+        if isinstance(data, dict) and "selections" in data:
+            data["selections"] = [
+                fix_selection(sel) for sel in data["selections"] if sel
+            ]
+        # Remove empty array fields from the top level
+        keys_to_delete = [k for k, v in data.items() if isinstance(v, list) and not v]
+        for k in keys_to_delete:
+            del data[k]
+        return data
+
+    def model_dump(self, *args, **kwargs):
+        data = super().model_dump(*args, **kwargs)
+        return self._post_process(data)
+
+    def model_dump_json(self, *args, **kwargs):
+        # Dump to python dict first, post-process, then dump to JSON
+        import json
+        from datetime import timezone
+        model_dump_kwargs = kwargs.copy()
+        json_kwargs = {}
+        # List of json.dumps kwargs you want to support
+        json_kwarg_names = ['indent', 'sort_keys', 'separators', 'ensure_ascii']
+        for key in json_kwarg_names:
+            if key in model_dump_kwargs:
+                json_kwargs[key] = model_dump_kwargs.pop(key)
+        # Get dict with Pydantic's processing (exclude_none, etc.)
+        data = super().model_dump(*args, **model_dump_kwargs)
+        data = self._post_process(data)
+        # Format timestamp as UTC RFC3339 string
+        if "timestamp" in data and isinstance(data["timestamp"], datetime):
+            utc_dt = data["timestamp"].astimezone(timezone.utc)
+            data["timestamp"] = utc_dt.strftime("%Y-%m-%dT%H:%M:%SZ")
+        return json.dumps(data, **json_kwargs)
 
 def main() -> None:
     print(


src/ssvc/selection.py
```diff
@@ -312,7 +312,51 @@ def model_json_schema(cls, **kwargs):
...
+            if key in model_dump_kwargs:
+                json_kwargs[key] = model_dump_kwargs.pop(key)
+        # Get dict with Pydantic's processing (exclude_none, etc.)
+        data = super().model_dump(*args, **model_dump_kwargs)
+        data = self._post_process(data)
+        # Format timestamp as UTC RFC3339 string
+        if "timestamp" in data and isinstance(data["timestamp"], datetime):
+            utc_dt = data["timestamp"].astimezone(timezone.utc)
+            data["timestamp"] = utc_dt.strftime("%Y-%m-%dT%H:%M:%SZ")
+        return json.dumps(data, **json_kwargs)
```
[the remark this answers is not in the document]
sei-vsarvepalli: The line L24 has 
`from datetime import datetime`

src/ssvc/selection.py
```diff
@@ -312,7 +312,51 @@ def model_json_schema(cls, **kwargs):
...
+        json_kwarg_names = ['indent', 'sort_keys', 'separators', 'ensure_ascii']
+        for key in json_kwarg_names:
+            if key in model_dump_kwargs:
+                json_kwargs[key] = model_dump_kwargs.pop(key)
+        # Get dict with Pydantic's processing (exclude_none, etc.)
+        data = super().model_dump(*args, **model_dump_kwargs)
+        data = self._post_process(data)
+        # Format timestamp as UTC RFC3339 string
+        if "timestamp" in data and isinstance(data["timestamp"], datetime):
+            utc_dt = data["timestamp"].astimezone(timezone.utc)
```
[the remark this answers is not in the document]
sei-vsarvepalli: But I want to always convert it to UTC. This looks like it will not.

src/ssvc/selection.py
```diff
@@ -312,7 +312,55 @@ def model_json_schema(cls, **kwargs):
...
+            selection["values"] = [v for v in list(values) if v]
+            return selection
+
+        # If this is a dict with selections, process each selection
+        if isinstance(data, dict) and "selections" in data:
+            data["selections"] = [
+                fix_selection(sel) for sel in data["selections"] if sel
+            ]
+        # Remove empty array fields from the top level
+        keys_to_delete = [k for k, v in data.items() if isinstance(v, list) and not v]
```
ahouseholder: Can we eliminate the need for the `keys_to_delete` loop by using `exclude_none=True` or `exclude_if = lambda: ...` in the class definition?

src/ssvc/selection.py
```diff
@@ -312,7 +312,55 @@ def model_json_schema(cls, **kwargs):
         schema = strip_nullable_anyof(schema)
 
         return order_schema(schema)
+    def _post_process(self, data):
+        """
+        Ensures all Selection.values are lists and removes empty array elements.
+        """
+        def fix_selection(selection):
```
ahouseholder: I see what this is doing, but I don't understand *why* we need it? Is there an example of a Selection object that needs this treatment? How would that get created? If so, can we put that into a unit test to demonstrate the need for this method?

src/ssvc/selection.py
```diff
@@ -312,7 +312,55 @@ def model_json_schema(cls, **kwargs):
         schema = strip_nullable_anyof(schema)
 
         return order_schema(schema)
+    def _post_process(self, data):
```
ahouseholder, line 315: Is this solving a problem on the serialization side when it might be more of an object validation problem? I would think we'd want to flag objects that look like what this method would fix on the validation side as the object is created rather than allowing bad objects to propagate until outbound serialization. But maybe I'm not understanding the problem.

src/ssvc/selection.py
```diff
@@ -312,7 +312,55 @@ def model_json_schema(cls, **kwargs):
...
+        return self._post_process(data)
+
+    def model_dump_json(self, *args, **kwargs):
+        # Dump to python dict first, post-process, then dump to JSON
+        import json
+        from datetime import timezone
+        model_dump_kwargs = kwargs.copy()
+        json_kwargs = {}
+        # List of json.dumps kwargs you want to support
+        json_kwarg_names = ['indent', 'sort_keys', 'separators', 'ensure_ascii']
```
ahouseholder: Why do we need to be picky about this instead of just passing `**kwargs` along? (If we do it this way, then we're breaking the expectation that model_dump_json passes `kwargs` through.)

src/ssvc/selection.py
```diff
@@ -312,7 +312,55 @@ def model_json_schema(cls, **kwargs):
...
+        for key in json_kwarg_names:
+            if key in model_dump_kwargs:
+                json_kwargs[key] = model_dump_kwargs.pop(key)
+        # Get dict with Pydantic's processing (exclude_none, etc.)
+        data = super().model_dump(*args, **model_dump_kwargs)
+        data = self._post_process(data)
+        # Format timestamp as UTC RFC3339 string
+        if "timestamp" in data and isinstance(data["timestamp"], datetime):
+            ts = data["timestamp"]
```
ahouseholder: Can we do this in a [field serializer](https://docs.pydantic.dev/latest/concepts/serialization/#field-serializers) instead? If so, I think this would let us avoid having to override `model_dump_json()` entirely.

src/ssvc/selection.py
sei-vsarvepalli: yeh will get rid of this for something simpler

`return json.dumps(data, **{k: v for k, v in kwargs.items() if k in json.dumps.__code__.co_varnames})`

src/ssvc/selection.py
sei-vsarvepalli: This was to standardize UTC conversion. I think it is not necessary. People may want to keep Timezone. I will get rid of it.

src/ssvc/selection.py
sei-vsarvepalli: The problem as stated is to remove any bloated empty arrays. As these may be noisy when introduced to larger records like CVE or CSAF record. I think we are also attempting to use if present should not be empty as requested by Thomas Schmidt for the schema, it may also require this empty arrays to be not present. For now to get this into CVE system, this tool may help for those wanting to generate proper CVE records but no noise.

PR #1006, opened 2025-10-09
Add API examples endpoints

### LINKED ISSUE #1005: Add `/examples/*` to basic FastAPI
For the object types listed below, provide endpoints in the basic FastAPI implementation such that:

`GET /examples/<object-type>` provides an example object of that type
`POST /examples/<object-type>` runs the provided JSON object through the pydantic validation for the object type

This will provide a simple way to bootstrap folks who want to develop code that emits valid (schema compliant) SSVC JSON objects.

The object types we need this for include:

- _decision points_
- _decision tables_
- _selection lists_
- Optional — because they're technically covered by their inclusion in the above, but we might as well do it anyway since it's not much extra effort
    - _decision point values_ — an individual element of a decision point value
    - _selection_ — an individual element of a selection list
    - _minimal decision point values_  — an individual element of an individual selection item

ahouseholder:
- resolves #1005 

This PR complements #1002. Starting it as a draft for discussion purposes.

Still TODO:

- [x] add tests for api endpoints
- [x] add readme to `src/ssvc/api`

---

This pull request introduces a new examples API router, improves the organization and documentation of the SSVC API, and refines serialization logic for several models. The main changes include adding endpoints for retrieving and validating example SSVC objects, updating the API documentation, and enhancing how falsy fields are handled in model serialization.

**API Improvements**
* Added a new `examples` router (`src/ssvc/api/v1/routers/examples.py`) that provides GET and POST endpoints for sample and validation of SSVC objects (Decision Points, Decision Tables, Selections, References, etc.), making it easier to test and understand the API.
* Registered the `examples` router in the main v1 API router (`src/ssvc/api/v1/routers/v1_router.py`), and updated the main API router registration to include a versioned prefix and tags for better organization. [[1]](diffhunk://#diff-eb864ec960b086a7acd8ab6a23080b48d768d222bb59aa74648f6098e262d432R29) [[2]](diffhunk://#diff-eb864ec960b086a7acd8ab6a23080b48d768d222bb59aa74648f6098e262d432L39-R41) [[3]](diffhunk://#diff-e479a95b1397d134f3f7c4fbb80c389b545d4b08c6c0af2b2dd82d786fe8a1adL48-R54)

**Documentation**
* Added a comprehensive README for the SSVC API (`src/ssvc/api/README.md`) with setup instructions for development and production environments.

**Example Data**
* Added a new module (`src/ssvc/examples.py`) containing reusable example objects for Decision Points, Decision Tables, Selections, and References, used by the examples API endpoints.

**Serialization Logic**
* Improved serialization of `Reference` and `SelectionList` models to automatically remove falsy fields when dumping to JSON, and ensured default values for missing fields, leading to cleaner API responses. [[1]](diffhunk://#diff-09b89f8c1d6ac82fe49f1a4939d82fd90767af95229434db7c5f3b79197bb8aaR136-R157) [[2]](diffhunk://#diff-09b89f8c1d6ac82fe49f1a4939d82fd90767af95229434db7c5f3b79197bb8aaR256-R260)
* Refactored and removed legacy post-processing methods in `SelectionList` to rely on the new model serializer approach, simplifying the codebase and improving maintainability.

Let me know if you want to walk through any specific endpoint or serialization logic in more detail!

### DIFF
--- a/src/test/test_selections.py
+++ b/src/test/test_selections.py
@@ -474,7 +474,7 @@ class MyTestCase(unittest.TestCase):
         )
         dumped = s.model_dump()
         with self.assertRaises(Exception):
-            del dumped["values"]
+            del dumped['values']
 
 
 if __name__ == "__main__":


src/ssvc/selection.py
ahouseholder, line 1: Overriding`model_dump` and `model_dump_json` works when you're interacting directly with the Python objects. But when you add FastAPI to the mix, it turns out that FastAPI tries to outsmart you by creating parallel class definitions on the return trip through how the `response_model` interacts with `jsonable_encoder`, which bypasses these overrides. So instead, a more universal solution is to add the `@model_serializer` decorated method here (which I also needed to handle missing/empty `Reference.summary` values too).

src/ssvc/api/main.py
```diff
@@ -45,12 +44,14 @@
     },
 )
 
-app.include_router(router_v1)
+app.include_router(router_v1, prefix="/ssvc/api/v1", tags=["SSVC API v1"])
```
ahouseholder, line 47: > [!NOTE]
> This changes the prefix for the whole API. If we don't want to force the `/ssvc/api` prefix (or we'd rather it just be `/ssvc/v1`), we should change this line.

ahouseholder: Also, instead of having `ssvc.api.v1.routers.v1_router.router` assign the `/v1` part of the prefix, this PR changes it so the `/v1` prefix is added by the `include_router` call. I'm not sure which way is preferable.

src/ssvc/selection.py
```diff
@@ -140,6 +141,28 @@ class Reference(BaseModel):
...
+
+        Args:
+            data: The input data dictionary.
+
+        Returns:
+            The modified data dictionary with summary set to an empty string if it was missing.
+
+        """
+        if "summary" not in data or not data["summary"]:
+            data["summary"] = ""
```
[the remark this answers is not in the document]
sei-vsarvepalli: This statement looks like confounded. It still keeps data["summary"] - I think we are trying to check that it is not None perhaps?

```
        elif not data["summary"]:
            data["summary"] = ""
            
 ```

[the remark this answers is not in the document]
sei-vsarvepalli: Hey @copilot perhaps you should us `data["summary"] is None` especially if you are being [nitpick]

src/ssvc/api/main.py
sei-vsarvepalli: no preference really. I think it is fine the way you have it.

PR #1121, opened 2026-04-24
Add ADR-0015: SSVC JSON schemas use SchemaVer for versioning

### LINKED ISSUE #601: Create ADR for SSVC JSON schema versioning
> Yeh - I think we can move that way later without a problem. The idea would be to provide "breaking" vs "non-breaking" Schema changes to be version controlled.  Mostly from [Snowplow's musings](https://docs.snowplow.io/docs/pipeline-components-and-applications/iglu/common-architecture/schemaver/) . 

> The challenge with some of these issues seems to show that much of the schema versioning itself is so experimental and we are trying to imagine an operational viability which is sort of impossible to know till we have enough usage.

_Originally posted by @sei-vsarvepalli in https://github.com/CERTCC/SSVC/pull/599#discussion_r1674439644_

This issue can be resolved by a PR that includes a new ADR addressing JSON schema versioning and pointing to the snowplow.io link in the above comment.

sei-renae:
This pull request documents a new architectural decision to adopt SchemaVer for versioning SSVC JSON schemas, and updates the ADR index accordingly. The main goal is to provide a versioning scheme tailored to the compatibility needs of JSON schema consumers, rather than relying solely on Semantic Versioning (SemVer), which is already used for other SSVC components.

**Documentation of new schema versioning approach:**

* Added ADR 0015 (`0015-ssvc-json-schemas-use-schemaver.md`) describing the adoption of SchemaVer for SSVC JSON schemas, detailing the rationale, versioning rules, and implications for compatibility and process.
* Updated the ADR index (`index.md`) to include the new decision record for SchemaVer usage.Resolves #601. Documents the decision to adopt SchemaVer (MODEL-REVISION-ADDITION) for SSVC JSON schema versioning, explaining why it fits schema evolution better than SemVer or CalVer, and defines the rules for incrementing each component.


sei-vsarvepalli (approved):
Approved .

PR #1160, opened 2026-06-10
Fix calculator checkbox order dependency in tree pruning

Skalman:
check_select() pruned the decision tree incrementally via the rendered DOM nodes at the toggled decision point's depth, so branches hidden by an earlier selection were never updated and reappeared with stale state when re-checked. Checkboxes therefore only behaved when toggled in decision-point order.

Rebuild the pruning for the whole tree from the complete checkbox state on every toggle, walking the tree data (saved _schildren) instead of the rendered nodes, so toggle order no longer matters.


sei-vsarvepalli (approved):
Looks good!

PR #1186, opened 2026-06-18
Pypi release can be now automated.

### LINKED ISSUE #1175: pypi publish with GitHub workflow.
Current Github to PyPi publish workflow is manual with keys and authentication with API. For more safe option use the GH workflow to stage a release when ready.

Current script is 
```
git checkout pypi
pypitag=$(date +%Y.%-m.%-d%H%M)
git tag $pypitag
git push origin pypi:pypi --tags
uv lock
uv sync
python -m build
twine check build/*
twine upload build/*
```

use GH to pypi integration and staged push of the pypi package to production pypi and development pypi locations.

sei-vsarvepalli:
Related to #1175  - The pypi branch requires that we modify pyproject.toml

To release a project to pypi
1. Ensure a Tag is created in claver like format currently `date +%Y.%-m.%-d%H%M` 
2. Push the updates to the branch pypi - then it will use OIDC authentication to publish the software to pypi

@ahouseholder - please make sure the pyproject.toml changes are okay with you. This is required to ensure that pypi package grabs the required files from "src/" directory.

### DIFF
--- a/.github/workflows/pypi_publish.yml
+++ b/.github/workflows/pypi_publish.yml
@@ -30,8 +30,6 @@ jobs:
 
       - name: Verify distributions
         run: |
-          pip install -U twine
-          twine check dist/*
           ls -lah dist/
 
       - name: Upload distributions
--- a/.github/workflows/pypi_publish.yml
+++ b/.github/workflows/pypi_publish.yml
@@ -1,9 +1,12 @@
+---
 name: Build and Publish to PyPI
 
 on:
   push:
     tags:
       - "20*"
+    branches:
+      - pypi
 
 permissions:
   contents: read
@@ -71,4 +74,3 @@ jobs:
         uses: pypa/gh-action-pypi-publish@release/v1
         with:
           verbose: true
-
--- a/.github/workflows/pypi_publish.yml
+++ b/.github/workflows/pypi_publish.yml
@@ -5,8 +5,7 @@ on:
   push:
     tags:
       - "20*"
-    branches:
-      - pypi
+
 
 permissions:
   contents: read
--- a/.github/workflows/pypi_publish.yml
+++ b/.github/workflows/pypi_publish.yml
@@ -30,7 +30,7 @@ jobs:
 
       - name: Verify distributions
         run: |
-          uv pip install twine
+          pip install twine
           twine check dist/*
           ls -lah dist/
 
--- a/.github/workflows/pypi_publish.yml
+++ b/.github/workflows/pypi_publish.yml
@@ -1,0 +1,74 @@
+name: Build and Publish to PyPI
+
+on:
+  push:
+    tags:
+      - "20*"
+
+permissions:
+  contents: read
+
+jobs:
+  build:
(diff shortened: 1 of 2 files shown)


.github/workflows/pypi_publish.yml
```diff
@@ -0,0 +1,75 @@
...
+  build:
+    name: Build distribution
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout source
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
```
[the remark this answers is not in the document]
sei-vsarvepalli: this is no longer needed. we are going to depend on the tag on ly primarily for pypi publishing.

.github/workflows/pypi_publish.yml
```diff
@@ -0,0 +1,63 @@
+name: Build and Publish to PyPI
+
+on:
+  push:
+    tags:
+      - "20*"
```
sei-ahouseholder, line 6: since this is a glob pattern and not a regex, should we use `"20??.*.*"` instead to ensure we don't match on non-calver patterns?

.github/workflows/pypi_publish.yml
```diff
@@ -0,0 +1,63 @@
+name: Build and Publish to PyPI
+
+on:
```
sei-ahouseholder, line 3: don't we need to also limit it to only pushes to the publish branch?

sei-ahouseholder (commented):
I think the original #1175 is underspecified. See my [comment](https://github.com/CERTCC/SSVC/issues/1175#issuecomment-4810488873) there on why I don't think `on push tags 20*` is the right way to go, but we need to decide on which of a few alternative paths to take from here.

ahouseholder (approved):

ahouseholder:
Following conversations with the team, we're good with merging this PR. There's a need for a follow-on issue to resolve how we're using CalVer (see [ADR-0013](https://github.com/CERTCC/SSVC/blob/a5e938dc598eda88256c5d23f8d2346b0f24d260/docs/adr/0013-ssvc-project-versions.md) because this seems to change what we need to do, but we haven't reached a full resolution on how we intend to increment versions yet. (I.e., are we going to "pure timestamp" CalVer like `2026.6.221541`, or do we stick with some concept of "major is always year, but we only change minor to represent a significant change -- so `2026.6.3` might actually come in July 2026 if it's just a small change from the `2026.6.0` release), but if it were a bigger release it would be `2026.7.0`. We've basically been operating under the assumption that only the `YYYY.M.0` release has to fall in the month indicated, and we can increment the patch release as needed between bigger `.0` releases. The pypi versions are using `YYYY.-m.-dHHMM` format though, so that contradicts ADR-0013. The resolution isn't necessarily to just comply with ADR-0013, it's to re-evaluate whether ADR-0013 is serving our needs now, and decide whether to supersede it with a new ADR or not.

@sei-vsarvepalli is planning to create a new issue to capture this concern for future consideration.

Issue #1124, opened 2026-05-08
Scale slider doesn't work on ssvc-calc Coordinator Triage


sei-renae:
**Describe the bug**
I can't adjust the scale of the Coordinator Triage model

**To Reproduce**

1. Go to [ssvc-calc](https://certcc.github.io/SSVC/ssvc-calc/)
2. Select Coordinator Triage from the dropdown
3. Try to use the slider to adjust scale
4. It doesn't select

**Expected behavior**
I can click+drag the slider the adjust scale

**Screenshots**

<img width="529" height="535" alt="Image" src="https://github.com/user-attachments/assets/0f430683-6da5-4201-b900-88a0a70cee69" />

**Platform details**
Firefox on macos tahoe

**Additional context**
The slider works on other decision models

sei-vsarvepalli:
In the web inspector or Javascript Console you can do, which will fix this. But I will get a PR out to fix it too.


`$('#zoomcontrol').css({'z-index':100}) `

Issue #1198, opened 2026-07-09
Wrong mouse-over text for "Supplier Cardinality" node in Coordinator Triage model


sei-mjaconski:
**Describe the bug**
In the Coordinator Triage model, mousing over Supplier Cardinality nodes shows the help text for Supplier Engagement. 

**To Reproduce**

1. Go to [ssvc-calc](https://certcc.github.io/SSVC/ssvc-calc/)
2. Change the Sample Decision Model to "Coordinator Triage (1.0.0)"
3. Locate any "Supplier Cardinality" node, and mouse over it. 
4. See that it shows the blue help text for "Supplier Engagement". 
5. Compare by mousing over an actual "Supplier Engagement" node, which uses the same help text.

**Expected behavior**
Mousing over a "Supplier Cardinality" node should show help text relating to cardinality decision information (Credible vs Not Credible). 

**Screenshots**
<img width="801" height="294" alt="Image" src="https://github.com/user-attachments/assets/d991e0ee-a3dd-4f4f-b4ba-06e8a67fc65c" />
(Unfortunately the snipping tool can't capture cursors, so just trust me that the node is being moused over here)

**Platform details**
Chrome on Windows

**Additional context**
N/A

sei-vsarvepalli:
This is fixed now with 2026.7.0 released today. Thanks @sei-mjaconski

PR #1200, opened 2026-07-09
Updates to fix zoomcontrol Issue #1124

sei-vsarvepalli:
Resolution to #1124 for the publish website to get the right update.

### DIFF
--- a/docs/ssvc-calc/findex.html
+++ b/docs/ssvc-calc/findex.html
@@ -144,7 +144,8 @@
 	  </table>
 	</p>
 	<section id="evaluate_section"></section>
-	<div id="zoomcontrol" class="graphy embedded d-none">
+	<div id="zoomcontrol" class="graphy embedded d-none"
+	     style="z-index:100">
 	  <div class="d-none">&#128269;</div>
 	  <input type="range" orient="vertical" alt="Zoom Graph"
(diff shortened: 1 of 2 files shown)


sei-vsarvepalli:
Also resolves the issue in #1198  - where there is a collision between div names if the strings are similar like "Supplier Cardinality" vs "Supplier Engagement"

sei-renae:
The scroll bar works now, but could there be a horizontal scroll as well? 
image

sei-vsarvepalli:
> The scroll bar works now, but could there be a horizontal scroll as well? image

The zoom works diagonally to do both really. It is not a horizontal only zoom per se. 

https://github.com/CERTCC/SSVC/blob/386564747de7551bd93a30e95cf45a3a18eb47d2/docs/ssvc-calc/ssvc.js#L2957

sei-renae:
image

I mean this scroll bar - can we get a horizontal one of these?

sei-vsarvepalli:
> image
> I mean this scroll bar - can we get a horizontal one of these?

not really. The SVG element growth is something div is unaware of as the display changes in the iframe.  To really work out the display the whole iframe should be abandoned and the display integrated into the mkdocs directly. That is a lot more effort!

ahouseholder (approved):

PR #1223, opened 2026-08-03
Initial commit of SSVC-AI decision point evaluation skill

sei-renae:
## Changes

- Adds a reusable `eval-against-decision_point` base skill that defines the shared SSVC evaluation workflow, confidence thresholding, dependency handling, and JSON output schema.
- Adds a `technical-impact` SSVC skill for evaluating the Technical Impact decision point against vulnerability reports.
- Adds OpenCode configuration for the `ssvc-docs` reference so skills can fetch canonical SSVC docs and source files at runtime.

## additional info for local testing
I have in my global `.config/opencode/.opencode.jsonc`:
```
"permission": {
    "webfetch": "ask",
    "bash": {
      "pip install*": "ask",
      "pip3 install*": "ask",
      "npm install*": "ask",
      "brew install*": "ask",
      "brew upgrade*": "ask",
      "curl*": "ask",
      "wget*": "ask",
      "*": "ask"
    }
  }
```
and in my global `.config/opencode/instructions.md`:
```
# Global Coding Instructions

## Python

- All Python code must be run inside a virtual environment (venv).
- Before running any `python` or `python3` command, ensure a venv is active or create one first.
- Never install Python packages globally (no `pip install` or `pip3 install` outside a venv, and never use `--break-system-packages`).
- To set up a venv for a project:
  ```bash
  python3 -m venv .venv
  source .venv/bin/activate
  pip install <package>```
```

### DIFF
--- a/.agents/skills/eval-against-decision_point/SKILL.md
+++ b/.agents/skills/eval-against-decision_point/SKILL.md
@@ -1,0 +1,208 @@
+---
+name: eval-against-decision_point
+description: Use ONLY when evaluating SSVC decision points against a vulnerability report. Provides shared evaluation process, JSON output schema, process of elimination, confidence threshold, and sub-skill invocation pattern used by all SSVC decision point skills.
+---
+
+# SSVC Base Evaluation Skill
+
+This skill defines the shared rules and process that all SSVC decision point evaluation skills follow. It is not invoked directly by a user — it is referenced by each individual decision point skill.
+
+---
+
+## Role
+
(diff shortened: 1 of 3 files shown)


ahouseholder (approved):
I think this is good enough to merge, assuming that it's framed as "we're experimenting here, and the fact that it's in `main` doesn't mean that we endorse it as ready for production use or fit for purpose.

sei-vsarvepalli (approved):

PR #1242, opened 2026-09-10
Add MIT License to agent skills

sei-renae:


sei-vsarvepalli:
Okay should also update https://github.com/CERTCC/SSVC/blob/main/LICENSE accordingly. I will do it later when I can sneak a little break..

sei-vsarvepalli (approved):
Useful to have this. I think .agent/skills/LICENSE is redundant but okay to leave it.

